PR #7, opened 2017-03-09
Describe the relocations that LLVM is currently using.

sunfishcode:
Because we want to be able to write C++ global addresses into the data section,
we can't count on using a get_global to read the value of an imported global to
get the C++ global address. In view of that, and the fact that it's also simpler
to write addresses directly into the code section than use an actual get_global,
drop the get_globals altogether. This results in several changes to the
relocations.

And, add an addend field to C++ global address relocations. I think this ends
up making sense because it makes sense for the addend to be signed, but some of
the relocations apply to fields which are unsigned, so storing the addend in them
is awkward.

LLVM isn't quite finished implementing the stack pointer part of this, and as I wrote this up, I also realized that it's missing support for relocating type indices in `call_indirect` instructions, however I'm posting it now so that we can start talking about the other parts.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -59,14 +59,15 @@ A relocation type can be one of the following:
   instruction, e.g. taking the address of a function.
 - `2 / R_WEBASSEMBLY_TABLE_INDEX_I32` - a function table index encoded as a
   [uint32], e.g. taking the address of a function in a static data initializer.
-- `3 / R_WEBASSEMBLY_GLOBAL_ADDR_LEB` - a global index encoded as a 5-byte
+- `3 / R_WEBASSEMBLY_MEMORY_ADDR_LEB` - a linear memory index encoded as a 5-byte
   [varuint32]. Used for the immediate argument of a `load` or `store`
   instruction, e.g. directly loading from or storing to a C++ global.
-- `4 / R_WEBASSEMBLY_GLOBAL_ADDR_SLEB` - a global index encoded as a 5-byte
+- `4 / R_WEBASSEMBLY_MEMORY_ADDR_SLEB` - a linear memory index encoded as a 5-byte
   [varint32]. Used for the immediate argument of a `i32.const` instruction,
   e.g. taking the address of a C++ global.
-- `5 / R_WEBASSEMBLY_GLOBAL_ADDR_I32` - a global index encoded as a [uint32],
-  e.g. taking the address of a C++ global in a static data initializer.
+- `5 / R_WEBASSEMBLY_MEMORY_ADDR_I32` - a linear memory index encoded as a
+  [uint32], e.g. taking the address of a C++ global in a static data
+  initializer.
 
 [varuint32]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varuintn
 [varint32]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varintn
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -39,47 +44,62 @@ A "reloc" section is defined as:
...
+  5-byte [varint32]. Used to refer to the immediate argument of a `i32.const`
+  instruction, e.g. taking the address of a function.
+- `2 / R_WEBASSEMBLY_TABLE_INDEX_I32` - a function table index encoded as a
+  [uint32].
+- `3 / R_WEBASSEMBLY_GLOBAL_ADDR_LEB` - a global index encoded as a 5-byte
+  [varuint32]. Used for the immediate argument of a `load` or `store`
+  instruction.
+- `4 / R_WEBASSEMBLY_GLOBAL_ADDR_SLEB` - a global index encoded as a 5-byte
+  [varint32]. Used for the immediate argument of a `i32.const` instruction,
+  e.g. taking the address of a function.
```
sbc100: taking the address of a global?

Linking.md
```diff
@@ -39,47 +44,62 @@ A "reloc" section is defined as:
...
+- `0 / R_WEBASSEMBLY_FUNCTION_INDEX_LEB` - a function index encoded as a 5-byte
+  [varuint32]. Used for the immediate argument of a `call` instruction.
+- `1 / R_WEBASSEMBLY_TABLE_INDEX_SLEB` - a function table index encoded as a
+  5-byte [varint32]. Used to refer to the immediate argument of a `i32.const`
+  instruction, e.g. taking the address of a function.
+- `2 / R_WEBASSEMBLY_TABLE_INDEX_I32` - a function table index encoded as a
+  [uint32].
+- `3 / R_WEBASSEMBLY_GLOBAL_ADDR_LEB` - a global index encoded as a 5-byte
+  [varuint32]. Used for the immediate argument of a `load` or `store`
+  instruction.
```
sbc100: e.g. reading/writing to a global?

Linking.md
```diff
@@ -39,47 +44,62 @@ A "reloc" section is defined as:
...
-  of a C global
-
-For relocation types other than `R_DATA` the following fields are present:
+- `0 / R_WEBASSEMBLY_FUNCTION_INDEX_LEB` - a function index encoded as a 5-byte
+  [varuint32]. Used for the immediate argument of a `call` instruction.
+- `1 / R_WEBASSEMBLY_TABLE_INDEX_SLEB` - a function table index encoded as a
+  5-byte [varint32]. Used to refer to the immediate argument of a `i32.const`
+  instruction, e.g. taking the address of a function.
+- `2 / R_WEBASSEMBLY_TABLE_INDEX_I32` - a function table index encoded as a
+  [uint32].
```
sbc100: I guess this is for function addresses stored in the data segment?  perhaps mention that?

Linking.md
```diff
@@ -39,47 +44,63 @@ A "reloc" section is defined as:
...
-relocation points to the offset of the padded immediate value within the code
-section, allowing the linker to both read the current value and write an updated
-one.
+Merging of globals sections requires re-numbering of the globals.
+
+This convention requires the first global in the global section to be a mutable
+i32 global initialized to "STACKTOP" in the "env" module. During linking, only
+one of these globals is kept, and it remains the first global. This is a
+simple convention which allows code to reference global `0` without needing to
+be rewritten.
```
sbc100: What about the fact that any imported globals come first in the index space of globals?  Are we relying on the fact that there will be no imported globals in the final linked executable?

sbc100:
Can/should we land this now?  I'd like to update wabt to support these changes.

Linking.md
```diff
@@ -91,6 +96,22 @@ present:
 | index  | `varuint32`      | the index of the global used        |
 | addend | `varint32`       | addend to add to the address        |
 
+For `R_WEBASSEMBLY_TYPE_INDEX_LEB` relocations the following fields are
```
sbc100: What are these needed for?

Linking.md
sbc100: Ignore me I saw your comment above

sunfishcode:
I'm still working on the stack pointer issue, but we can merge it now and continue to iterate if that works for you.

sbc100:
Yes, lets merge this now and iterate.  lgtm

sbc100:
Can you explain how and why the linker would make use of the new R_WEBASSEMBLY_STACK_POINTER?   Is it so that the stack point of each object can be merged into a single global in the output?

sunfishcode:
Merging, as discussed above. We'll continue to iterate.

Yes, the purpose of R_WEBASSEMBLY_STACK_POINTER is so that the stack pointers of each object can be merged. I'm open to other ideas too.

Issue #23, opened 2017-11-10
Linking ABI: import the memory and table?


sunfishcode:
Currently, clang emits code that imports the stack pointer, but doesn't import the `memory` or `table`. It doesn't export the ones it defines either. Clearly the linker can still do what it needs, but it would be nice to have some consistency here. Is it best to import the `memory` and `table`, or do something else?

dschuff:
I guess if we want to maintain the "object files are runnable" idea then we probably need to locally define the stack pointer (along with memory and table) and make the linker more magical. Otherwise it probably makes sense to have object files just import everything.
Even in that case (i.e. the case where we give up on saying object files are runnable) it probably makes sense to have some consistent idea of how valid or how close to runnable object files are? E.g. we could at least say "object files are valid wasm" (does that currently hold?). If we do that then we'd want to locally define memory and table (and make it shared if needed, etc) to keep that property.

sbc100:
We do already say that they should be valid wasm:  "These modules are not expected to be directly executable and have certain constraints on them, but are otherwise well-formed WebAssembly modules".  

 I don't have strong opinions about whether llvm-generated object should import or declare their memories and tables, either one seems reasonable, and does not make the more or less runnable does it?

sunfishcode:
I agree. I'm now less interested in making .o files directly runnable than I initially was. But I am still interested in consistency where it makes sense, and this seems like an area where it makes sense.

guybedford:
I haven't followed the most recent updates on the web assembly stack pointer conventions, so this may be completely off track, but I do think it's worth considering (or ruling out, and if not already) that a nice property to have might be the ability to merge multiple web assembly binaries into a single binary (a sort of composibility optimization for better network and parse times in the final production module graph, like we have with JS modules).

For example say a trend is to publish small wasm + JS packages to npm and we want Webpack users to be able to bundle that up for production. Usually when we have lots of separate JS files we want to combine them into common JS files for production, and ES modules have this composibility property for us. If the common wasm binaries could be squashed into a single common wasm binary that might be a nice optimization to have as well. This arbitrary composibility just kind of appeals as a concept.

[...]

PR #26, opened 2017-12-06
Add a comment about the motivation for WASM_DATA_SIZE.

sunfishcode:
Add new linking metadata types for memory and tables.

Also, remove WASM_DATA_SIZE, which is redundant with the "minimum" field
of the linear memory description.

This documents the changes proposed in LLVM patch [D40875](https://reviews.llvm.org/D40875).

I'm not attached to most of the particular details here; my main interest is in switching to imports for memory and tables.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -145,6 +145,9 @@ The current list of valid `type` codes are:
 
 - `2 / WASM_DATA_SIZE` - Specifies the total size of the static data used by the
   module, including both initialized and zero-initialized (bss) data.
+  Note that this is similar to the "minimum" field in the linear memory
+  description, however it's in units of bytes, so it's not limited to being
+  a multiple of the page size.
 
 - `3 / WASM_DATA_ALIGNMENT` - Specifies the alignment of the data section.  This
   tells the linking what constraints are placed on the location of the data


sunfishcode:
Also, for reference, this addresses #23.

sbc100:
I think you still need WASM_DATA_SIZE since wasm memory is specified in page size and we want objects to be able to has less than a whole page of global data.

Regarding the other changes?  Why not just say that only a single memory and table is allowed in linkable objects?  Seems like it keeps things simpler for now.

sbc100:
Regarding WASM_DATA_SIZE, I was thinking of removing it also, but for other reasons.   It basicaly came into existence to handle BSS, but we probably need to re-address how bss works, especially now that the linker is only including segments that are explicitly referenced by symbols.

sunfishcode:
Ah, good point about WASM_DATA_SIZE not being limited to page size multiples. I'll update the patch to add it back in, and if you have other ideas, we can go from there.

Regarding the other changes, another option is to leave those out and omit `WASM_STACK_POINTER` entirely. It looks like lld is currently just looking for the name `__stack_pointer` anyway. What would you think of that?

sunfishcode:
Thinking about this more, I think `WASM_STACK_POINTER` was more important when the stack pointer wasn't an import. Making it an import gives it a name, which we can use instead.

That makes switching to imports for linear memory and tables also nice, because it gives them names too.

sbc100:
Yes.. I think we can drop WASM_STACK_POINTER

sbc100:
lgtm, but you should update the change description.

PR #31, opened 2017-12-08
Add COMDAT linking data encoding

NWilson:
This is the documentation for the corresponding change in https://reviews.llvm.org/D40844, to be documented as and when that format is accepted.

See the discussion in LLVM bugzilla and Phabricator for information on COMDAT.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -150,7 +150,7 @@ The current list of valid `type` codes are:
   tells the linking what constraints are placed on the location of the data
   section in the final binary.
 
-- `6 / WASM_COMDAT_INFO` - Specifies the COMDAT groups of associated linking
+- `7 / WASM_COMDAT_INFO` - Specifies the COMDAT groups of associated linking
   objects, which are linked only once and all together.
 
 For `WASM_SYMBOL_INFO` the following fields are present in the
@@ -222,7 +222,10 @@ and where a `ComdatSym` is encoded as:
 | kind     | `varuint32`    | Type of symbol, one of:                     |
 |          |                |   * `0 / WASM_COMDATA_DATA`, a data segment |
 |          |                |   * `1 / WASM_COMDATA_FUNCTION`             |
-| index    | `varuint32`    | Index of the symbol in the Wasm object      |
+| index    | `varuint32`    | Index of the data segment or function in    |
+|          |                | the Wasm module (depending on kind).  The   |
+|          |                | function index is in the Wasm index space,  |
+|          |                | and must not reference an import.           |
 
 
 Merging Global Section
--- a/Linking.md
+++ b/Linking.md
@@ -150,6 +150,9 @@ The current list of valid `type` codes are:
   tells the linking what constraints are placed on the location of the data
   section in the final binary.
 
+- `6 / WASM_COMDAT_INFO` - Specifies the COMDAT groups of associated linking
+  objects, which are linked only once and all together.
+
 For `WASM_SYMBOL_INFO` the following fields are present in the
 subsection:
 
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -194,6 +197,33 @@ subsection:
...
+| name_str    | `bytes`      | UTF-8 encoding of the name                |
+| count       | `varuint32`  | Number of `ComdatSym` in `comdat_syms`    |
+| comdat_syms | `ComdatSym*` | Sequence of `ComdatSym`                   |
+
+and where a `ComdatSym` is encoded as:
+
+| Field    | Type           | Description                                 |
+| -------- | -------------- | ------------------------------------------- |
+| kind     | `varuint32`    | Type of symbol, one of:                     |
+|          |                |   * `1 / WASM_COMDATA_DATA`, a data segment |
```
sunfishcode: I'm unclear on the meaning of `WASM_COMDAT_DATA`. In my current understanding, segments don't have their own names; they're essentially attached to globals (in the ABI), and globals are entities that do have names. So then when we want them to participate in COMDAT, it seems like it would be sufficient to just include the globals in the group, and not the segments. Would that make sense?

Linking.md
```diff
@@ -194,6 +197,33 @@ subsection:
...
+| comdats | `Comdat*`   | Sequence of `Comdat`
+
+where a `Comdat` is encoded as:
+
+| Field       | Type         | Description                               |
+| ----------- | ------------ | ----------------------------------------- |
+| name_len    | `varuint32`  | length of `name_str` in bytes             |
+| name_str    | `bytes`      | UTF-8 encoding of the name                |
+| count       | `varuint32`  | Number of `ComdatSym` in `comdat_syms`    |
+| comdat_syms | `ComdatSym*` | Sequence of `ComdatSym`                   |
```
sunfishcode: It might be a good idea to add an unused flags field here, in case we want to later add things like ODR checking, same-size checking, "pick the biggest one", or other features from other COMDAT systems.

Linking.md
NWilson: Good question. It originally seemed to make the implementation a bit simpler, but you're right I shouldn't need both.

In the LLVM MC layer, it's the section (MCSection) that has the Comdat associated with it, and that's what ELF encodes. I think it is common in general a section/segment to contain multiple globals, and the frontend just makes sure to use the Comdat name as part of the section key, to ensure that each section/segment only accumulates symbols from the same Comdat.

If we're following ELF and just encoding the MCSection structures, I think it should be the data segments only that we mark as Comdat, and not the global symbols. LLD already finds the matching data segment for each global when it puts it in the Symtab, so we can get the Comdat from there.

That should be most general, since each global must have a data segment, but there's not quite as strong a requirement in the other direction, as data segments can be referenced via relocations without using a global symbol.

NWilson:
Ping. Could this be merged now that the LLVM change has been merged? Thanks.

Linking.md
```diff
@@ -150,6 +150,9 @@ The current list of valid `type` codes are:
   tells the linking what constraints are placed on the location of the data
   section in the final binary.
 
+- `6 / WASM_COMDAT_INFO` - Specifies the COMDAT groups of associated linking
```
sbc100: This ended up being 0x7 in the final commit.

Linking.md
```diff
@@ -194,6 +197,33 @@ subsection:
...
+| comdat_syms | `ComdatSym*` | Sequence of `ComdatSym`                   |
+
+and where a `ComdatSym` is encoded as:
+
+| Field    | Type           | Description                                 |
+| -------- | -------------- | ------------------------------------------- |
+| kind     | `varuint32`    | Type of symbol, one of:                     |
+|          |                |   * `0 / WASM_COMDATA_DATA`, a data segment |
+|          |                |   * `1 / WASM_COMDATA_FUNCTION`             |
+| index    | `varuint32`    | Index of the symbol in the Wasm object      |
```
sbc100: Wasm function index or data segment index depending `kind`.

sbc100:
Sorry those first two review comments were old (please ignore).

Linking.md
NWilson: Done. I used the term "Wasm index space" in the updated text - which is true already, since right now that's the only index space that exists, and it will stay true once we add a symbol index space.

PR #39, opened 2018-01-22
Create an explicit symbol table

NWilson:
This implements the spec changes discussed in issue https://github.com/WebAssembly/tool-conventions/issues/38.

Changes:
* Gets rid of globals as a way of modelling data
* Add globals for real Wasm globals such as `__stack_pointer`
* Specifies ways that relocation processing can/must fail
* Specifies the Wasm object file symbol table


NWilson:
@sbc100 Ready to merge now it's all in LLD!

PR #42, opened 2018-02-28
Remove DATA_SIZE and DATA_ALIGNMENT

sbc100:
They are no longer needed to used in llvm/lld now that we
have segment info and symbol table metadata.


sbc100:
@NWilson

PR #44, opened 2018-02-28
Use uint8 consistently for enum types

sbc100:
this is inline with the offical spec where value types are specified as single bytes and
not LEBs.

Also, use lower case when naming elements of the binary structure.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -55,7 +55,7 @@ A `relocation_entry` begins with:
 
 | Field    | Type                | Description                    |
 | -------- | ------------------- | ------------------------------ |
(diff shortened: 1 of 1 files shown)


sbc100:
@NWilson, do you want to join the CG so that I can add you as a reviewer on this kind of thing?  (You will also be able to work directly in this repo too).

NWilson:
I'd like to, but I'm not sure I can - I don't think my employer will support me as a corporate member, and the CG needs an employer declaration for corporate members, and won't accept personal members if they're employed at a company with an interest in the technology. I just don't get much work time allowance for it.

PR #57, opened 2018-04-25
Add version to linking metadata section

sbc100:
This change only adds single version to the linking medadata section.  Add more detailed
information such as ABI and ABIFlags can be considered a separate change.

See https://github.com/WebAssembly/tool-conventions/issues/54


NWilson:
Makes sense to merge, since it's ready to go into LLVM.

PR #73, opened 2018-11-08
Linking.md: Add note about start section

sbc100:


Linking.md
```diff
@@ -386,3 +386,17 @@ adjusted to reflect new offsets in the code section.
...
+By default the static linker should not output a WebAssembly start section.
+
+Rational: We considered using the WebAssembly start function for running static
+constructors and/or the main entry point to the program.  However, in practice
+we found that allowing arbitrary code in the start section is problematic as the
+module exports not yet available to the embedder while running this code.  A
+common example is the module memory itself. If the code in the start function
+wants to transfer any data to the embedder (e.g. printf) this will not work as
+the embedded cannot yet access the modules memory.  This extends to embedder
+functions that call back into module-exported function.
```
kripken: Might be worth mentioning the `__post_instantiate` stuff in dynamic linking in this context?

sbc100:
I wasn't aware that the ESM proposal was doing that.   It seems a little strange that the start function behaviour would be different in the module case.  

In any case there is no plan to do this for the non-module case right?  I'm happy to put a note here saying that we may use the start function in the future if exports are available prior to running it, but I don't see how we can specific that today.. since they are not.    Are we committed to having the exports available when start runs?   That seems like it would require a significant spec change.

Linking.md
```diff
@@ -386,3 +386,17 @@ adjusted to reflect new offsets in the code section.
 adjusted to reflect new offsets in the combined sections.
 
 [names_sec]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#name-section
+
+Start Section
+-------------
+
+By default the static linker should not output a WebAssembly start section.
+
+Rational: We considered using the WebAssembly start function for running static
```
littledan: Nit: "Rationale"

littledan:
@sunfishcode My latest draft at https://github.com/WebAssembly/esm-integration/pull/21 does not make the exports available when the start function is running--it leaves the "instantiation" step in the JS API in tact. This differs from the current webpack semantics.

Linking.md
sbc100: I might leave that as a separate change.    We haven't yet talked about how to call __wasm_call_ctors/main/__post_instantiate.

xtuc:
> And Webpack et al may polyfill this by converting the start function into an export when needed

That's correct, Webpack does it unconditionally.

---

Webpack does wrire up the export at compile time, but that's to work around the current ESM semantics, e.g. a wasm module importing another wasm module's export. Webpack's runtime does only provide a trampoline.

sbc100:
Any more thoughts on this wording?

PR #79, opened 2018-11-21
Rename R_WEBASSEMBLY_TYPE_INDEX_LEB

aheejin:
This renames R_WEBASSEMBLY_TYPE_INDEX_LEB to
R_WEBASSEMBLY_SIGNATURE_INDEX_LEB, to be clearer and more readable.


rossberg:
Drive-by comment: the Wasm spec intentionally calls this type index, not signature index. The reason being that we expect additional forms of type definitions to appear in the type section in the future, see e.g. the GC proposal.

aheejin:
Thanks! I guess it's better to not change this then.

PR #84, opened 2018-12-15
Specify relative orders of custom sections

aheejin:
This specifies relative orders of
- "linking" custom section and the data section
- "reloc." custom sections and the "linking" custom section


sbc100:
Perhaps "must" rather than "should"?

PR #94, opened 2019-01-30
Add Webpack/webassemblyjs in producers

xtuc:
While `Webpack` is not a SDK, it conforms with the description
> designates the top-level "thing" that the developer installs and interacts with directly to produce the wasm module

For ref https://github.com/webpack/webpack/issues/8429

`webassemblyjs` is the underlying tool that Webpack uses.


xtuc:
Could someone please merge this? If I need consensus I'm happy to bring that up during a WebAssembly GC meeting.

PR #99, opened 2019-02-20
Target features section

tlively:
Adds a new "target_features" custom section that lists the features an
object was built with. It is an error to link objects with mismatched
target feature sets.

The presence of the "atomics" feature also replaces the value of the
shared bit on object file memory import as the signal to the linker
that it should use shared memory. The target features section is more
generally useful and less hacky than the old solution.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -34,6 +34,14 @@ to: e.g. "reloc.CODE" for code section relocations.  However everything after
 the period is ignored and the specific target section is encoded in the reloc
 section itself.
 
+The linker additionally checks that linked object files were built targeting
+compatible feature sets. Unlike native targets, WebAssembly has no runtime
+feature detection, and the presence of unsupported features causes a binary to
+fail to validate. It is therefore almost always a user error to link together
+object files built for different feature sets, and the linker can be helpful by
+enforcing the invariant that the feature sets must be consistent by checking
+objects' target features sections.
+
 Relocation Sections
 -------------------
 
@@ -285,6 +293,32 @@ and where a `comdat_sym` is encoded as:
 |          |                |   * `3 / WASM_COMDAT_EVENT`                 |
 | index    | `varuint32`    | Index of the data segment/function/global/event in the Wasm module (depending on kind). The function/global/event must not be an import. |
 
+Target Features Section
+-----------------------
+
+The target features section is an optional custom section with the name
+"target_features". The target features section must come after the
+["producers"](#linking-metadata-section) section.
+
+The body of the target features section is a vector of unique strings, each
+representing a feature of the target. The generally accepted feature strings
+are:
+
+1. `atomics`
+2. `bulk-memory`
+3. `exception-handling`
+4. `nontrapping-fptoint`
+5. `sign-ext`
+6. `simd128`
+
+The "atomics" feature string is special: if it is present, the linker will
+produce a binary that uses a shared memory.
+
+It is an error if two object files in a single link contain different sets of
+strings in their target features sections. Objects with no features section may
+be linked with any other objects.
+
+The linker does not emit a features section in its output.
 
 Merging Global Sections
 -----------------------
@@ -371,17 +405,6 @@ which reference a data symbol.
 
 Segments are linked as a whole, and a segment is either entirely included or
 excluded from the link.
-
-Merging Memory Sections
------------------------
-
-It is an error to link together object files with shared and unshared
-memories. Object files with unshared memories are produced by tools that are not
-aware of WebAssembly threads or are configured to produce code for a
-single-threaded environment. These objects may have been compiled from
-thread-aware source code but had their atomic operations stripped, so it would
-be dangerous to allow users to accidentally link these objects together with
-properly thread-aware objects.
 
 Merging Custom Sections
 ----------------------
--- a/Linking.md
+++ b/Linking.md
@@ -311,11 +311,12 @@ The body of the target features section is a vector of entries:
 The recognized prefix bytes and their meanings are below. When the user does not
 supply a set of allowed features explicitly, the set of allowed features is
 taken to be the set of used features. Any feature not mentioned in an object's
-target features section is not used by that object.
+target features section is not used by that object, but is not necessarily
+prohibited in the final binary.
 
 | Prefix     | Meaning |
 | ---------- | ------- |
-| 0x00       | This object uses this feature, and the link fails if this feature is not in the allowed set. |
+| 0x2b (`+`) | This object uses this feature, and the link fails if this feature is not in the allowed set. |
 | 0x2d (`-`) | This object does not use this feature, and the link fails if this feature is in the allowed set. |
 | 0x3d (`=`) | This object uses this feature, and the link fails if this feature is not in the allowed set or if any object does not use this feature. |
 
--- a/Linking.md
+++ b/Linking.md
@@ -37,10 +37,11 @@ section itself.
 The linker additionally checks that linked object files were built targeting
 compatible feature sets. Unlike native targets, WebAssembly has no runtime
 feature detection, and the presence of unsupported features causes a binary to
-fail to validate. It is therefore almost always a user error to link together
-object files built for different feature sets, and the linker can be helpful by
-enforcing the invariant that the feature sets must be consistent by checking
-objects' target features sections.
+fail to validate. It is therefore important for the user to have explicit
+control over the features used in the output binary and for the linker to
+provide helpful errors when instructed to link incompatible or disallowed
+features. This feature information is stored in a custom ["target feature
+section"](#target-features-section).
 
 Relocation Sections
 -------------------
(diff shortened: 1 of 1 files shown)


tlively:
This section could be represented in the .s format either as normal data (like the producers section is) or by a new `.target_features` directive that takes flags corresponding to features.

tlively:
Expanding a bit on the problems this feature solves:

Most immediately, we need a way for the linker to determine whether to use a shared or unshared memory in its output. The linker should use a shared memory if any of its input objects are thread-aware. The current plan for determining whether an object is thread-aware is to set the shared bit on its imported memory, but assigning implicit meaning to a pre-existing bit in the wobject that the linker has no other reason to consider is rather hacky. A much cleaner solution would be to explicitly state that the wobject is thread-aware in some metadata section. This is what the target features section provides.

But why does the linker need to detect whether to use a shared memory? Couldn't it just continue using the `--shared-memory` flag it has today? This brings us to a more general problem.

Currently emscripten users need to pass feature flags both at compile time and at link time. Emscripten uses the link time flags to tell the linker whether to use a shared memory and to tell binaryen what features it is allowed to use in optimization. Forcing the user to repeat these flags at link time violates the DRY principle and the principle of least astonishment for users used to native linkers.

If link time feature flags served some other purpose, such as allowing the linker to to select library versions based on the desired output feature set, then they would be very useful. But currently it is an error in the best case and silently useless in the worst case to provide at link time any set of feature flags besides the union of all compile time feature flags, making them an unnecessary complication.

The target features section both improves usability now by removing duplicated flags and is also an important step toward more feature-aware toolchain usability improvements in the future.

CC @kripken

xtuc:
Great idea!

It would be useful to add some JavaScript features too:

- `mutable-global`; this would be very useful for the transformation done by Webpack. We avoid using the feature because of no proper way to feature test.

- `bigint`; the compiler can check whatever exporting to JS funcs with a signature containing i64 types if supported.

---

In pratice I think that the same version of a brower can enable/disable different features on different plateform, that sounds to me like a new fingerprinting technique. I'm not an expert, I just wanted to bring that up.

tlively:
The more I think about this, the more I think it makes sense to move the feature-compatibility testing out of the linker and into the toolchain driver. This would make the linker more similar to a feature-agnostic native linker and allow different toolchains to have different feature compatibility policies. The linker would still know how to merge the target features sections from multiple object files together, though.

Taking this design change to the extreme also means removing the logic for determining whether to use a shared or unshared memory from the linker and keeping the current `--shared-memory` linker flag. Different toolchains could have different policies, but would still be able to observe the target features sections in making their decision.

@xtuc I don't think it would make sense for LLVM to emit information about JavaScript features into the target features section, since it generally doesn't know anything about the JavaScript glue. That being said, other parts of toolchains could certainly store JS-related information in the target features section as appropriate. It's just a list of strings, after all.

Regarding fingerprinting, I would expect all toolchains to strip the target features section before emitting the final binary, since it's not meant to be used by the browser or any other consumer of binaries. So I don't think fingerprinting or information leakage is an issue here.

tlively:
This PR has been updated to reflect all the great conversations I had with various people today, especially with @jgravelle-google, @kripken, and @dschuff. Below are a number of design questions that came up and our answers to them. @sunfishcode I'm particularly interested in whether this design meets the needs of the Rust toolchain.

__1. Should the target feature section contain features allowed at compilation or features actually used in the object?__
It should contain all features allowed at compilation, since the features actually used in the object can already be inferred from the object contents but the allowed features cannot already be inferred.

__2. Should the linker do compatibility checking or should some other tool do the checking?__
The linker should do the checking because many toolchains do not want to take a dependency on other tools, and it is nicer to keep the linker as the only tool that reads multiple object files at once.

__3. Are there any special features or are they all treated the same?__
The linker checks for the presence of the "atomics" feature to determine whether or not to emit a shared memory, but all features are treated equivalently for the purpose of checking object compatibility. This keeps the compatibility mechanism simpler and easier to reason about.

__4. Should users be able to specify allowed target features at link time?__
Although we don't want to require specifying features at link time in the common case, it is important to allow users to explicitly specify the allowed features at link time so they know exactly what is coming out the other side.

__5. Should the encoded features be only strings or should they contain compatibility policy information as well?__
If the features are only strings, the only possible compatibility policy that treats all objects and features the same is exact feature set matching. This does not support common usage patterns such as a single SIMD-enabled object being linked with non-SIMD objects, so we want richer semantics built in to the target feature data. These semantics will be encoded as prefix bytes before the feature strings.

_6. What prefix bytes should there be?_
The prefixes are given in the text of the PR. These are some example use cases:


| Prefix | Use case |
| ------ | -------- |
| null (0x00) | A SIMD-enabled object file may use this prefix with the "simd128" feature to signal that it uses SIMD but is able to link with objects that do not use SIMD. |
| '-' (0x2d)  | An object that has had all of its atomics stripped during codegen might use "-atomics" to prevent itself from being linked with other multithreaded objects. |
| '=' (0x3d)  | This may be used to enforce that all object files share an ABI or to enforce that all object files have all the same features. |


__7. Should the compatibility policy be configurable by linker flags?__
No, the prefix bytes should be enough for toolchains to encode any compatibility policy they want. More prefix bytes may be added as necessary in the future.

__8. Should the linker use feature information to determine what to include in the link?__
Although it would be great for the linker to be able to select between multiple versions of libraries or functions based on the set of allowed features, this is a problem that is being deliberately punted to the future. A future design for this problem should be able to use the target feature section as a building block, for example by associating different versions of a function with different target feature sections in a fat binary.

tschneidereit:
Also CC @fitzgen and @alexcrichton for feedback regarding Rust toolchain needs.

sunfishcode:
This sounds great, and lines up with my intuition. But I have some questions about 6. :-)

Why a NUL byte? That a little surprising since the others are printable ASCII.

Do I understand the 4 states?
 "-foo" - this module is incompatible with foo being used
 "\0foo" - this module uses foo
 "=foo" - this module uses foo, and requires all other modules to declare the same
 <not present> - producing compiler didn't know about the "foo" feature

If so, what is the difference between <not present> and "-foo"?

alexcrichton:
Ah thanks for the cc, and for Rust this sounds pretty good! One use case we try to cater to (like with native platforms) is to ship a binary standard library which is conservative in its usage of features (e.g. doesn't use simd128) but allows user code compiled-on-demand to use it as much as possible. In that manner we'd want user code to be able to use `simd128` and such features, and that'd be compatible with a libstd that doesn't use `simd128`.

Having the ability to blanket reject a feature I think also makes sense because our standard library is compiled quite differently if `atomics` is enabled, so the current standard library would want to fail to link if anything else uses atomics, and we'd provide our own mechanism for fixing that.

I'd also have a similar question as @sunfishcode to clarify the meaning of "nothing was mentioned". I think one way to interpret it is the same as `-foo` ("incompatible with and nothing else can use it") but I think it could also be interpreted as "this module doesn't use the feature and everything else is allowed to use it". The latter interpretation is somewhat more flexible but less future-proof.

tlively:
> This sounds great, and lines up with my intuition. But I have some questions about 6. :-)
> 
> Why a NUL byte? That a little surprising since the others are printable ASCII.

That's reasonable. The other option I was thinking of was `+`, which is perhaps more consistent.

> Do I understand the 4 states?
> "-foo" - this module is incompatible with foo being used
> "\0foo" - this module uses foo
> "=foo" - this module uses foo, and requires all other modules to declare the same

Right.

> - producing compiler didn't know about the "foo" feature

Not necessarily. The compiler may know about "foo", but not use it and not want to prohibit linking with it. For example, an object that does not use SIMD but is ok with being linked with SIMD-enabled objects would simply not mention SIMD.

> If so, what is the difference between `<not present>` and "-foo"?

"No opinion" vs "Prohibited in link"

Under the current design, it is impossible to prohibit linking against unknown future features because they are in the `<not present>` category by necessity. If the Rust toolchain would like to be able to prohibit linking against unknown features, we could add an extra catch-all policy byte to the section. Possible values would be "allow unknown features" and "prohibit unknown features". We would also add a new prefix `?` meaning "unused, but not prohibited" to distinguish between unused but allowed features and unknown but prohibited features.

@alexcrichton Would Rust want the ability to set this catch-all policy for unknown features?

alexcrichton:
Nah I don't think we'd have any desire one way or another for unknown features. If necessary we can explicitly list a compatibility mode for all known features for each compilation and then we'd just update that list for new features as they come out. So in that sense we could work either way! (I was largely just curious)

tlively:
Ok, for simplicity I think we can safely leave out the catch-all mechanism then, unless someone else chimes in with a use case.

tlively:
I updated the prefixes to use '+' instead of NUL and clarified in the text that unmentioned features are not necessarily disallowed in the output. Does anyone have any remaining questions or concerns about this proposal?

sunfishcode:
This sounds good to me. I'm a little concerned about interactions between the various prefixes, however we can work out the details as we proceed.

tlively:
Thanks for your feedback, everyone! I'm going to merge this and hopefully implement it next week. I'm also happy to discuss further if anyone does have any more concerns or usecases. @alexcrichton and @fitzgen, it would be great if you could send me the details of the policy you would like to use in the Rust toolchain so I can add it as an option in the LLVM backend, unless you'd rather handle it as a postprocessing step on the object files.

alexcrichton:
Oh if we can get native LLVM support that'd be much nicer than postprocessing object files on our end, we always feel bad doing that!

For our toolchain I suspect the only thing we'll really do is that whenever the `atomics` feature is activated we'd want to specify `=atomics`, but otherwise we'd just have `+foo` for all other features.

Issue #93, opened 2019-01-29
Emscripten and the Producers Section


kripken:
After a lot of consideration, I don't think we want to change Emscripten to emit the producers section by default in release builds. Posting this issue to note that and explain why.

 * It would be surprising if `emcc` started one day to emit information like what compiler and tools our users use - most people would probably be surprised by that, and some might be concerned about it, for privacy or security reasons.
 * There is no precedent for this in the Web space: minifiers don't emit a `// minified by $MINIFERNAME` comment in your minified code. I can't even find a flag to do this optionally in any minifiers.
   * Oddly there *is* precedent in the native space, as clang and gcc do emit the compiler name and version, but after much digging I never found a good reason for this (just a 20 year old comment about [SVR4 compatibility](https://gcc.gnu.org/ml/gcc-patches/1999-01/msg00455.html)). Most links online about this are users that find out about it and are surprised and sometimes unhappy (about code size, privacy, various bugs that it causes), and many projects end up [stripping](https://github.com/Alexpux/MINGW-packages/issues/21#issuecomment-340286825) it [out](https://cs.chromium.org/chromium/src/build/config/compiler/BUILD.gn?q=fno-ident&sq=package:chromium&dr=C&l=1903).
 * When users ask `emcc` to emit the smallest binary, we should do that, unless there's a very strong reason, and I'm not sure users would agree that the metrics we are talking about here are reason enough, based on discussions I had with some of them. The metrics are mostly of interest to browser vendors and tool creators, but it's the users that ship the extra bytes. Many users will probably flip a flag to remove those bytes, if they heard about that flag's existence, which suggests it should be on by default.

These are reasons for specifically not emitting the producers section in `emcc` by default. We could add an option for users that do want to do so, and of course other tools may have different factors to consider (in particular, Emscripten is used by ordinary developers, while tools like LLVM, wabt, or binaryen are used by toolchain developers, so the considerations might be different).

lukewagner:
I can see the logic behind these reasons and I wish they had been brought up in [earlier](https://github.com/WebAssembly/tool-conventions/issues/63) [discussions](https://github.com/WebAssembly/meetings/blob/master/2018/CG-10-02.md#there-is-some-positive-discussion-on-an-issue-in-the-tool-conventions-repo), where I was specifically hoping to hear from tooling people whether the producer section would actually show up in release builds.

[...]

PR #100, opened 2019-02-21
Update Producers Section to match #93

lukewagner:
As discussed in #93 and the recent [CG meeting](https://github.com/WebAssembly/meetings/blob/master/2019/CG-02-19.md#update-on-producers-section-tool-conventions93-points-to-likely-non-use-in-practice-and-so-next-steps-would-be-to-either-remove-or-soften-wording).

### DIFF
--- a/ProducersSection.md
+++ b/ProducersSection.md
@@ -1,33 +1,10 @@
 # Producers Section
 
-The purpose of the producers section is to provide an optional,
-highly-structured record of all the distinct tools that were used to produce
-a given WebAssembly module. A primary purpose of this record is to allow
-broad analysis of toolchain usage in the wild, which can help inform both wasm
-producers and consumers.
-
-The producers section is a
-[custom section](https://webassembly.github.io/spec/core/binary/modules.html#custom-section)
-and thus has no semantic effects and can be stripped at any time.
-Since the producers section is relatively small, tools are encouraged to emit
-the section or include themselves in an existing section by default, keeping
-the producers section even in release builds.
-
-WebAssembly consumers should avoid using the producers section to derive
-optimization hints. To ensure portable performance, hints should be
-standardized in a separate custom section, probably in the core spec's
-[Custom Sections appendix](https://webassembly.github.io/spec/core/appendix/custom.html).
-
-An additional goal of the producers section is to provide a discrete, but
-easily-growable [list of known tools/languages](#known-list) for each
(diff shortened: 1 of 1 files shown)


alexcrichton:
I personally feel like the phrasing of "size-optimizing toolchains" isn't quite right here because the Rust toolchain, for example, is (as always) size optimizing yet we intend to continue to emit the producers section by default.

There are other examples of "useless in production" sections being emitted by default by toolchainS such as the `name` section and dwarf debugging information. These sections (and the producers section) are all trivially stripped at any point in the build process, and we're always going to ensure there's documentation for standard tools (like `wasm-strip` in wabt) to strip these sections as well as CLI flags for our own tooling to strip the sections (like `--keep-debug`, `--remove-producers-section`, and `--remove-name-section`).

It seems like it'd be unfortunate to get into a situation where everyone's code-golfing how small their wasm can be to the point where very useful information is being stripped for a proportionally larger loss of functionality.

ProducersSection.md
```diff
@@ -1,33 +1,10 @@
...
-However, wasm consumers may provide less accurate telemetry results for unknown
-names or even emit diagnostics encouraging the name to be put on the known list.
-
-Since version information is useful, but highly-variable, each field value
-is accompanied with a version string so that the name can remain stable
-over time without requiring frequent updates to the known list.
+The producers section provides an optional, structured way to record all of the
+distinct tools that were used to produce a given WebAssembly module. The
+producers section is a [custom section](https://webassembly.github.io/spec/core/binary/modules.html#custom-section)
+and thus has no semantic effects. Size-optimizing toolchains are likely to
```
alexcrichton: Here some phrasing I might personally prefer is:

> Standard tooling like `wasm-strip` will remove the producers section, and toolchains which emit the producers section by default are recommended to have an option to remove it as well.

lukewagner: Thanks, makes sense.

kripken:
I feel like the size comment had value. I think it depends on which level of the tools we focus on. That is, I can see how Rust may not want to make any decision at all about the producers section, and just leave it to later tools. For a complete toolchain though, like say emcc, which does not assume any tool will run after it, it does make sense to strip the producers section by default for size (and information leakage) reasons.

lukewagner:
Happy to let someone else take over this PR and put the wording just right from a producer's perspective.

PR #104, opened 2019-04-03

sbc100:


sunfishcode:
Linkind.md describes a format we've [declared to be stable](http://releases.llvm.org/8.0.0/docs/ReleaseNotes.html#changes-to-the-webassembly-target), however the PR here adds bits from the dynamic linking ABI which is not currently stable (["Note: there is no stable ABI here yet."](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md)).

Concerning whether it should be stabilized in its current form, I have concerns about [PIC code being incompatible with static linking](https://reviews.llvm.org/D60049), with [the use of the `env` module and a one-level namespace](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md#imports), and with the format [not being tested outside of Emscripten](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md#implementation-status).

Is there a way you can add these relocations to wabt without committing to them as stable features?

sbc100:
Sure.  I share all of those concerns too.   I'll add these to DynamicLinking.md instead.

Also, we should update the intro paragraph on this document so reflect that fact that its a real thing now.

sbc100:
Closed in favor of #106

PR #107, opened 2019-04-16
Document shared memory and bulk-memory interaction

tlively:


Linking.md
```diff
@@ -504,3 +507,33 @@ extends to all embedder functions that might want to call back into the module.
 If some future version of the WebAssembly spec allows for module exports to be
 available during execution of the start function it will make sense to
 reconsider this.
+
+
+Shared Memory and Passive Segments
+----------------------------------
+
+When shared memory is enabled, all data segments will be emitted as [passive
```
lars-t-hansen: I would perhaps go for "... enabled, data segments can be emitted ..." since there's no rule that says you must do this.

Linking.md
```diff
@@ -504,3 +507,33 @@ extends to all embedder functions that might want to call back into the module.
...
+will be emitted into a synthetic function, `__wasm_init_memory` that is called
+from `__wasm_call_ctors` before any constructors run. To make these
+`memory.init` instructions valid, a [DataCount section][datacount_section] will
+also be emitted.
+
+Note that `memory.init` and the DataCount section are features of the
+bulk-memory proposal, not the atomics proposal. Despite this, they are emitted
+by the linker in the presence of shared memory (which requires the `atomics`
+feature) whether or not the `bulk-memory` feature is enabled. This is because
+they are necessary to avoid re-initializing memory on each thread. Any toolchain
```
lars-t-hansen: "necessary" overstates the case.

Linking.md
```diff
@@ -504,3 +507,33 @@ extends to all embedder functions that might want to call back into the module.
 If some future version of the WebAssembly spec allows for module exports to be
 available during execution of the start function it will make sense to
 reconsider this.
+
+
+Shared Memory and Passive Segments
+----------------------------------
+
+When shared memory is enabled, all data segments will be emitted as [passive
+segments][passive_segments] to prevent each thread from re-initializing
```
sbc100: We should probably mention WebWorker here.  They are the reason we need to do this.

tlively: Done.

tlively:
This change in the linker is imminent, so it would be good to get this documentation in. I removed a bunch of text that is no longer relevant now that all the threads-supporting engines I know of (V8 + SpiderMonkey) also support bulk memory.

PR #108, opened 2019-04-19
WebAssembly SIMD Intrinsic API Documentation

rrwinterton:
This is the markdown description of the WebAssembly SIMD Intrinsic API's

### DIFF
--- a/wasmsimdinst.md
+++ b/wasmsimdinst.md
@@ -4,14 +4,16 @@
 1. [Introduction](#introduction)
 2. [Proposed Vector Types](#proposed-vector-types)
 3. [Load and Store Definitions](#load-and-store-definitions)
-4. [v128 wasm_v128_constant](#v128-wasm_v128_constant)
-5. [
+5. [Broadcast, Extract and Replace Definitions](#broadcast-extract-and-replace-definitions)
+6. [Arithmetic Operations Definition](#arithmetic-operations-definition)
+7. [Logical Operations Definition](#logical-operations-definition)
+8. [Future Single Instruction Multiple Data Instructions](#future-single-instruction-multiple-data-instructions)
 
 ## **Introduction** <a name=introduction></a>
-This document is a draft proposal for a list of WebAssembly Intrinsic Instructions for C/C++ Clang compiler and an explanation of some of the design decisions and current state of WebAssembly Intrinsics.
+This document is a proposed list of WebAssembly Intrinsic Instructions for C/C++ Clang compiler and an explanation of some of the design decisions and current state of WebAssembly Intrinsics.
 
 ##### **Proposed Vector Types** <a name=proposed-vector-types></a>
-The following is a proposed vector type.  In asking others I thought it could be defined like llvm does  i8x16  for i8x16 and u8x16 for u8x16 but there was a preference expressed for using the v notation so it was kept.
+The following is a proposed vector type.  In asking others I thought it could be defined like llvm does  i8x16  for i8x16 and u8x16 for u8x16 but there was a preference expressed for using the v notation for loads and stores so it was kept.
 
 typedef v128 i8x16 __attribute__((__vector_size__(16)));
 typedef int8_t i8x16 __attribute__((__vector_size__(16)));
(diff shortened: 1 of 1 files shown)


alexcrichton:
FWIW this all roughly matches what we're doing in Rust to expose SIMD intrinsics ([documented here](https://doc.rust-lang.org/nightly/core/arch/wasm32/)), although there's two main differences:

* We're dropping the `wasm_` prefix proposed here and instead scoping it in a `wasm` module (I think the prefix makes sense for C/C++ though)
* We're not currently distinguishing between types like `i8x16` and `f32x4`, but rather there's only one `v128` type which everything takes. Given that all the underlying instructions operate over a single `v128`, I wonder if it'd be best for C/C++ to reflect the same and take `v128` everywhere?

tlively:
@alexcrichton I was thinking that having separate types might be helpful for users. I don't think it's very important that WebAssembly's underlying type system be mirrored exactly at the source level, even for intrinsics. On the other hand, clang by default silently allows conversions between vector types of the same size, so users have to opt-in to getting any benefit from the separate types. Perhaps having separate types is not worth the extra effort. I would be interested in hearing more opinions about this.

alexcrichton:
Makes sense! We originally used separate types awhile back in Rust as well. I think that the most important capability to enable here is what wasm actually exposes, which is passing any vector anywhere. We do that in Rust by taking `v128` everywhere (so the types all match up), but it sounds clang enables the same behavior with implicit conversions. 

In that sense I don't think the type names in C/C++ would matter too much since the underlying functionality shouldn't change too much. This header file, though, will likely be used to model other languages which may be more strict about types than Clang specifically is today. For example we won't be able to follow this specification in Rust due to the lack of implicit type coercions. In that sense it seems like it'd be nice to have languages exposing SIMD for wasm be consistent amongst themselves!

wasmsimdinst.md
```diff
@@ -0,0 +1,611 @@
...
+
+~~~
+f32x4 wasm_f32x4_replace_lane(f32x4 a, imm i, float b)
+~~~
+Replaces the float element specified by the immediate i with the value of float b
+
+##### **Arithmetic Operations Definition** <a name=arithmetic-operations-definition></a>
+
+~~~
+i8x16 wasm_i8x16_add(i8x16 a i8x16 b)
```
Maratyszcza: missing comma

Maratyszcza:
I suggest that `wasm_v128_load` and `wasm_v128_store` should use generic pointers (`[const] void*`) rather than pointers to `v128` type. This is preferable because applications are likely to store data in arrays of standard data types (e.g. `float`, `uint8_t`, or `complex<float>`), and having pointer casts everywhere would make codes using WASM intrinsics less readable. If loads/stores operate with generic (`void*) pointers, these casts are not needed as any pointer can be implicitly cast to `[const] void*`.

gnzlbg:
@Maratyszcza IIUC, the main problem I see with `v128*` in the loads and stores is that WASM supports unaligned loads and stores (the alignment is part of the `memarg` immediate) but this API only supports aligned loads and stores because dereferencing a `v128*` is only valid if it is properly aligned for `v128`. 

> and having pointer casts everywhere would make codes using WASM intrinsics less readable. 

I don't think this is just an inconvenience. Casting a `float*` to a `v128*`, and passing it to these APIs, would invoke undefined behavior if the `float*` is not aligned to a multiple of the alignment of `v128` (because the API does not support unaligned loads and stores).

tlively:
Thanks for your comments, everyone! Let's move conversation to the [implementing PR](https://github.com/emscripten-core/emscripten/pull/8559) for now to avoid duplication. Once the implementation is merged we can circle back and fix up the documentation to match.

sunfishcode:
@tlively The Emscripten work has landed; is this PR ready to go now?

Maratyszcza:
@sunfishcode The documentation in this PR is very out-of-date

tlively:
Indeed. Some minimal docs for the intrinsics can be found here: https://clang.llvm.org/doxygen/wasm__simd128_8h.html. Emscripten also has more detailed usage docs for porting existing SIMD code: https://emscripten.org/docs/porting/simd.html.

PR #112, opened 2019-05-31
Update DynamicLinking.md with current status

sbc100:
This change removes the `Planned Changes` section and makes this
the primary ABI.   The old ABI is mentioned only in reference to the
emscripten implementation.

### DIFF
--- a/DynamicLinking.md
+++ b/DynamicLinking.md
@@ -1,8 +1,8 @@
 WebAssembly Dynamic Linking
 ===========================
 
-This document describes a proposed ABI for dynamic linking of WebAssembly
-modules along with the current implementation status.
+This document describes the current WebAssembly dynamic linking ABI used by
+emscripten and by the llvm backend when targeting emscripten.
 
 Note: This ABI is still a work in progress.  There is no stable ABI yet.
 
--- a/DynamicLinking.md
+++ b/DynamicLinking.md
@@ -168,6 +168,6 @@ In order to use the llvm output in emscripten (which still uses the old ABI
 described above) a binaryen pass is run as part of `wasm-emscripten-finalize`
 then coverts from the ABI descirbed in this document to the ABI used by the
 emscripten dynamic linker is somewhat different.  The emscripten ABI uses
-specially named functions (e.g. `g$foo` and `fp$foo`) for the accessing the
+specially named functions (e.g. `g$foo` and `fp$foo`) for accessing the
 addresses of the dynamically linked symbols rather than using WebAssembly
 globals.
--- a/DynamicLinking.md
+++ b/DynamicLinking.md
@@ -1,10 +1,10 @@
 WebAssembly Dynamic Linking
 ===========================
 
-This document describes the early plans for how dynamic linking of WebAssembly
-modules might work.
+This document describes a proposed ABI for dynamic linking of WebAssembly
+modules along with the current implementation status.
 
-Note: there is no stable ABI here yet.
+Note: This ABI is still a work in progress.  There is no stable ABI yet.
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
This doesn't appear to perform two-level namespacing. This is a feature which we've envisioned wanting, under the belief that it will simplify and speed up the runtime linking code, and going by the experience of people we've talked to familiar with both Mach-O (which has two-level namespacing) and ELF (which doesn't, except in rare cases).

And, this looks like it might still have the property that -fPIC code requires -shared, and can't be linked into a static executable. Is that the case? 

Neither of these is necessarily a show stopper, though I'd like to understand the tradeoffs involved here.

sunfishcode:
Another question about dynamic linking in wasm is that existing software dynamic library boundaries are often language-specific, such that it may be desirable to distinguish between dynamic library symbol visibility and wasm module symbol visibility. In the system described here, global symbols with "default" visibility are made into wasm exports. However, another possible approach would be to reserve wasm exports for things which are specifically designed as such, possibly with the help of bindings generator tools, so that languages could have their dynamic library APIs work however they want, but still export a clean wasm-bindings-powered API to the rest of the wasm module world.

This is admittedly a big topic, but I'd be interested in thoughts about whether the design proposed here could evolve into a system like this.

DynamicLinking.md
```diff
@@ -170,15 +164,10 @@ See `test_dylink_*` adnd `test_dlfcn_*` in the test suite for examples.
...
-
 In order to use the llvm output in emscripten (which still uses the old ABI
 described above) a binaryen pass is run as part of `wasm-emscripten-finalize`
-then coverts to the old ABI.  This pass will convert all `GOT.mem` and
-`GOT.func` entries into regular (non-imported) globals and sets there values by
-calling the `g$foo` functions during `__post_instantiate` before any other code
-is run.
+then coverts from the ABI descirbed in this document to the ABI used by the
+emscripten dynamic linker is somewhat different.  The emscripten ABI uses
+specially named functions (e.g. `g$foo` and `fp$foo`) for the accessing the
```
kripken: `for the` => `for`?

DynamicLinking.md
```diff
@@ -170,15 +164,10 @@ See `test_dylink_*` adnd `test_dlfcn_*` in the test suite for examples.
...
+addresses of the dynamically linked symbols rather than using WebAssembly
+globals.
```
kripken: Is this intended to change?

sbc100: yes, we should be able to convert emscripten to natively support the new ABI.  I am thinking of waiting for fastcomp deprecation to avoid supporting 2 ABIs in emscripten at the same time.

sbc100:
I've updated this PR to be even more specific about the fact this is currently emscripten-only.

@sunfishcode to answer your specific concerns:

Regarding the two level namespace.  I agree this has not yet be investigated sufficiently.  This current proposal is the flat namespace.  I'll open an issue to discuss the future use of a two level namespace.

Regarding linking `-fPIC` code into regular executable.  There is nothing in this proposal that specifically prevents this.  It happens that the current tools don't support it.  I agree that it is a common enough pattern that we probably should allow it.  I'll open a bug regarding that too.

Regarding separating wasm exports from language-level linking.  It sounds like you are suggesting that we try to avoid wasm imports/exports for language-level symbol resolution?  What other mechanism could a language employ for accessing symbols in other DLLs (if not imports/exports).   Its seems that the two level namespace, along with naming conventions should be enough here.  I agree that "export to the wider world" and "export to other DLLs in the same program" might need to be separate concepts.

Hopefully this PR is good to land because its just documenting the status quo.

PR #113, opened 2019-06-07

sbc100:


sunfishcode:
Is this different from #104, such that my comment in https://github.com/WebAssembly/tool-conventions/pull/104#issuecomment-479318674 applies here as well?

sbc100:
Doh! Re-uploaded an old branch I guess.  Sorry about that.

PR #116, opened 2019-07-16
Define thread local storage conventions

quantum5:


Linking.md
```diff
@@ -528,3 +528,43 @@ threads needs to support both of these proposals.
...
+
+To use thread-local storage properly, a thread should do the equivalent of the
+following pseudo-code upon startup:
+
+```
+global.get __tls_size
+if
+  global.get __tls_size
+  call malloc
+  call __wasm_init_tls
```
tlively: This might be more readable as an s-expression.

Linking.md
```diff
@@ -528,3 +528,39 @@ threads needs to support both of these proposals.
...
+  variables, this function will do nothing. Otherwise, the memory block will be
+  initialized with the passive segment `.tdata` via the `memory.init` instruction.
+  Additionally, `__tls_base` will be set to the value of the pointer argument.
+
+To use thread-local storage properly, a thread should do the equivalent of the
+following pseudo-code upon startup:
+
+```
+(if (global.get __tls_size) (then
+  (call __wasm_init_tls (call malloc (global.get __tls_size)))))
```
sbc100: Out of interest how are we arranging for this to get called on thread startup in emscripten?   Are we exporting it?

Linking.md
```diff
@@ -528,3 +528,39 @@ threads needs to support both of these proposals.
 
 [passive_segments]: https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#design
 [datacount_section]: https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#datacount-section
+
+Thread Local Storage
+--------------------
+
+Currently, only the `local-exec` model of thread local storage is defined.
```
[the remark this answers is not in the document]
quantum5: This comes from the ELF TLS models, which LLVM also uses. It's the model that only supports TLS in the main executable.

Linking.md
quantum5: The emscripten side of things is in https://github.com/emscripten-core/emscripten/pull/8976.

Linking.md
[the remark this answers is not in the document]
tlively: Ulrich Drepper's TLS document uses "local exec" in normal typeface, so we should probably adopt that style here.

Linking.md
```diff
@@ -528,3 +528,42 @@ threads needs to support both of these proposals.
...
+In a threaded build, the module will contain:
+
+* an immutable global variable of type `i32` called `__tls_size`.
+  Its value is the total size of the thread local block for the module,
+  i.e. the sum of the sizes of all thread local variables plus padding.
+  This value will be `0` if there are no thread-local variables.
+* a mutable global `i32` called `__tls_base`. This will be initialized to
+  `0` by default. When a thread is created (including the main thread), memory
+  will be allocated and this global will be initialized with the help of the
+  function `__wasm_init_tls`.
```
sunfishcode: The phrases "will contain", "will be `0`", "will be initialized" and "will be allocated" here make it a ambiguous who's responsible for doing what, and when. It's clarified somewhat in the description of `__wasm_init_tls` below, but the text up to this point would be easier to follow if it were either more explicit about who's responsible for doing what, or alternatively, if it said less about how things are used here and just explained it all below.

Linking.md
```diff
@@ -528,3 +528,42 @@ threads needs to support both of these proposals.
 
 [passive_segments]: https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#design
 [datacount_section]: https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md#datacount-section
+
+Thread Local Storage
+--------------------
+
+Currently, thread-local storage is only supported in the main WASM module
+and cannot be accessed outside of it. This corresponds to the ELF local
+executable TLS model.
```
sunfishcode: The ELF documentation refers to it as "local exec", so it's fine to calling "local executable" here, but we should add something like "aka local exec" somewhere so that it's clear that we mean the same thing.

PR #119, opened 2019-07-22
Add conventions for TLS alignment

### LINKED ISSUE #118: TLS data doesn't indicate an alignment
Currently the description of [thread local storage](https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md#thread-local-storage) allows specifying the size of the TLS section to the runtime to allocate, but it doesn't specify the alignment with which to allocate it. While a 4 or 8-byte alignment is probably good enough for most programs, I know that in Rust we've historically run into issues where SIMD things got into LTO builds which required a bigger alignment (like 16 bytes or so).

Would it be possible for a global like `__tls_align` be synthesized next to `__tls_size` to indicate to the runtime at what alignment it needs to be allocated at?

cc @quantum5

quantum5:
Fixes #118.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -549,6 +549,9 @@ In a threaded build, the linker will create:
   Its value is the total size of the thread local block for the module,
   i.e. the sum of the sizes of all thread local variables plus padding.
   This value will be `0` if there are no thread-local variables.
+* an immutable global variable of type `i32` called `__tls_align`.
+  Its value is the alignment requirement of the thread local block. Its value
+  is `0` if there are no thread-local variables.
 * a mutable global `i32` called `__tls_base`, with a `i32.const 0` initializer.
 * a global function called `__wasm_init_tls` with signature `(i32) -> ()`.
 
@@ -556,7 +559,10 @@ To initialize thread-local storage, a thread should do the equivalent of the
 following pseudo-code upon startup:
 
     (if (global.get __tls_size) (then
-      (call __wasm_init_tls (call malloc (global.get __tls_size)))))
+      (call __wasm_init_tls
+        (call memalign
+          (global.get __tls_align)
+          (global.get __tls_size)))))
 
 `__wasm_init_tls` takes a pointer argument containing the memory block to use
 as the thread local storage block of the current thread. It should do nothing if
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -549,14 +549,20 @@ In a threaded build, the linker will create:
   Its value is the total size of the thread local block for the module,
   i.e. the sum of the sizes of all thread local variables plus padding.
   This value will be `0` if there are no thread-local variables.
+* an immutable global variable of type `i32` called `__tls_align`.
+  Its value is the alignment requirement of the thread local block. Its value
+  is `0` if there are no thread-local variables.
```
alexcrichton: According to [this man page at least](https://linux.die.net/man/3/memalign) the `memalign` function requires a power-of-two alignment which 0 may violate (and rather it should be 1), so for the example below to work I wonder if this should default to 1 instead of 0?

sunfishcode: This should specify that the alignment is in units of bytes (rather than the log2). Also, we should document that the `__tls_size` is required to be a multiple of `__tls_align`, and that the value of `__tls_align` is a power of 2.

quantum5: If there are no thread-local variables, `__tls_size` will also be 0. The `memalign` call is guarded by an if and will not run if there is no TLS.

tlively: I see that `aligned_alloc` requires its size argument to be a multiple of its alignment argument, and it would save the runtime effort if the linker did the rounding, but what if an implementation wanted to save memory by using a different allocation function that has no such restriction. Do we want to allow for that? Is there a deeper reason than `aligned_alloc`'s requirements to require that `__tls_size` be a multiple of `__tls_align`?

sunfishcode: Nothing super deep; mostly the reasons you mention. Another weak reason is that C structs also have this restriction, so if we ever wanted to describe the TLS block as a C struct, it'd be useful. So I don't have a strong opinion here.

If the linker sorts the TLS entries, it should be able to avoid wasting significant space in most cases. I guess you can always construct pathological cases, like a TLS block which only contains one small field with a large alignment. My assumption was that that should be pretty uncommon.

tlively: I think I err on the side of giving maximum information and control to the runtime, so I would rather not have the linker round up the size. We should still use `aligned_alloc` in the example and note the rounding requirement, though.

sunfishcode: Works for me!

sunfishcode:
We should still require the alignment to be a power of 2 though.

PR #124, opened 2019-08-20
Allocate magic constants for reference-types proposal

Keno:
The reference types proposal adds support for multiple table indices,
requiring us to reason about tables in linking. To make sure the various
tools use consistent names/constants for the required new relocation
type, add it to the list here, giving it an authoritative number.

This is based on my WIP support for LLVM here:
https://reviews.llvm.org/D66035, but I figured it'd be good to get
this out of the way first, so we can simultaneously start adding
support to other tools and make sure everybody has consistent numbers.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -25,6 +25,7 @@ tasks need to be performed:
 - Merging of function sections (re-numbering functions)
 - Merging of globals sections (re-numbering globals)
 - Merging of event sections (re-numbering events)
+- Merging of table sections (re-numbering tables)
 - Merging of data segments (re-positioning data)
 - Resolving undefined external references
 
@@ -105,6 +106,10 @@ A relocation type can be one of the following:
 - `10 / R_WASM_EVENT_INDEX_LEB` - an event index encoded as a 5-byte
   [varuint32]. Used for the immediate argument of a `throw` and `if_except`
   instruction.
+- `13 / R_WASM_TABLE_NUMBER_LEB` - a table number encoded as a 5-byte
+  [varuint32]. Used for the table immediate argument in the table.*
+  instructions.
+
 
 [varuint32]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varuintn
 [varint32]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varintn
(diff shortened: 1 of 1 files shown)


Keno:
cc @aheejin

Linking.md
```diff
@@ -105,6 +106,10 @@ A relocation type can be one of the following:
 - `10 / R_WASM_EVENT_INDEX_LEB` - an event index encoded as a 5-byte
   [varuint32]. Used for the immediate argument of a `throw` and `if_except`
   instruction.
+- `13 / R_WASM_TABLE_NUMBER_LEB` - a table number encoded as a 5-byte
+  [varuint32]. Used for the table immediate argument in the table.*
+  instructions.
```
aheejin: Unrelated to this CL, but we are missing 11 and 12 here, where the [code](https://github.com/llvm/llvm-project/blob/303bbc3be28747c8b6e25fb697c1ba6c47b962aa/llvm/include/llvm/BinaryFormat/WasmRelocs.def#L16-L17) has 11 and 12. Do you think we should add 11 and 12 here too, @sbc100?

Linking.md
```diff
@@ -487,6 +494,9 @@ which case there is no legal value that can be written as the target of any
 weak events which may not be defined; a definition or import must exist for all
 event symbols in the linked output.)
 
+`R_WASM_TABLE_NUMBER_LEB` relocations may fail to be processed, in which
+case linking fails. This occurs if there is a weakly-undefined table symbol.
```
aheejin: Maybe for consistency with other entries, we can add something like

> in which case there is no legal value that can be written as the target of any `table.get` or `table.set` instruction. (This means the frontend must not generate weak tables which may not be defined; a definition or import must generate weak tables which may not be defined; a definition or import must exist for all table symbols in the linked output.)

Linking.md
```diff
@@ -215,15 +220,16 @@ where a `syminfo` is encoded as:
...
 
 For functions, globals and events, we reference an existing Wasm object, which
-is either an import or a defined function/global/event (recall that the operand of a
+is either an import or a defined function/global/event/table (recall that the operand of a
 Wasm `call` instruction uses an index space consisting of the function imports
 followed by the defined functions, and similarly `get_global` for global imports
 and definitions and `throw` for event imports and definitions).
 
-If a function, global, or event symbol references an import, and the
+If a function, global, event or table symbol references an import, and the
```
sbc100: Perhaps we should rephrase this as "If a symbols refers to an import, and the..."

Linking.md
```diff
@@ -265,7 +271,7 @@ The current set of valid flags for symbols are:
   Hidden symbols are not to be exported when performing the final link, but
   may be linked to other modules.
 - `0x10 / WASM_SYM_UNDEFINED` - Indicating that this symbol is not defined.
-  For function/global/event symbols, must match whether the symbol is an import
+  For function/global/event/table symbols, must match whether the symbol is an import
```
sbc100: How about "For non-data symbols.."

Linking.md
sbc100: Hmm.. I seem to remember that rather than specify this stuff for some symbol types we chose to make it illegal to have weak symbols of certain types.   Can you see any use for WEAK being applied to table or event symbols?

aheejin: Not really, because currently event symbols are only generated from `throw` and `br_on_exn` instructions, and there will be only one symbol for each language (e.g. C++) and that's generated by the compiler. For tables, I'm not sure we can declare weak tables. How can we declare weak globals?

Keno:
Updated with the wording changes, and I've removed the language on weak table symbols. We can figure out what language to add about weak symbols later. Sorry for the delay on this - I only have time to spend on this intermittently.

PR #127, opened 2019-09-06
Update passive memory initialization scheme

tlively:
Reduces requirements for runtimes by changing `__wasm_init_memory` to
be the start function rather than being called from
`__wasm_call_ctors` or manually. Also adds language about
synchronization guarantees provided by `__wasm_init_memory`.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -531,8 +531,9 @@ instantiation. `__wasm_init_memory` shall perform any synchronization necessary
 to ensure that no thread returns from instantiation until memory has been fully
 initialized, even if a module is instantiated on multiple threads
 simultaneously. This synchronization may involve waiting, so in a web context
-the runtime must ensure that the main thread does not race with worker threads
-to initialize memory. To make the `memory.init` and `data.drop` instructions
+the runtime must ensure that the module is instantiated either first on the
+browser's main thread without racing with worker threads or not at all on the
+browser's main thread. To make the `memory.init` and `data.drop` instructions
 valid, a [DataCount section][datacount_section] will also be emitted.
 
 Note that `memory.init` and the DataCount section are features of the
--- a/Linking.md
+++ b/Linking.md
@@ -525,7 +525,7 @@ memory. In a web context, using active segments would cause memory to be
 reinitialized every time the module is instantiated on a new WebWorker as part
 of spawning a new thread. The `memory.init` instructions that initialize these
 passive segments and the `data.drop` instructions that mark them collectible
-will be emitted into a synthetic function, `__wasm_init_memory` that is made the
+will be emitted into a synthetic function `__wasm_init_memory` that is made the
 WebAssembly start function and called automatically on instantiation but is not
 exported. `__wasm_init_memory` shall perform any synchronization necessary to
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -524,10 +524,16 @@ segments][passive_segments] to prevent each thread from reinitializing
...
-will also be emitted.
+passive segments and the `data.drop` instructions that mark them collectible
+will be emitted into a synthetic function, `__wasm_init_memory` that is made the
+WebAssembly start function and called automatically on
+instantiation. `__wasm_init_memory` shall perform any synchronization necessary
+to ensure that no thread returns from instantiation until memory has been fully
+initialized, even if a module is instantiated on multiple threads
+simultaneously. This synchronization may involve waiting, so in a web context
+the runtime must ensure that the main thread does not race with worker threads
+to initialize memory. To make the `memory.init` and `data.drop` instructions
```
aheejin: Don't we prevent races in wasm code? Does the runtime still need to do something?

tlively: Yes, the wasm code has the synchronization. The runtime only needs to prevent the main thread from waiting on the main thread in a web context. How can I make that clearer in the text?

alexcrichton:
Without having a build on hand, my guess for the new scheme would be that the `start` function reserves an atomic counter in memory, the first thread performs memory initialization, all threads perform `memory.drop`, and then all threads use `atomic.notify` and `i32.atomic.wait` to wait for the "initialized" counter to get set to finished?

This does sort of seem like an unfortunate gotcha that you cant race instantiation with the main thread, but I'm not sure there's really a better scheme other than that...

tlively:
@alexcrichton Yes, example implementation [here](https://github.com/llvm/llvm-project/commit/09768c5d7abb398e92ed5fd1660d0ea69976943a#diff-8b1f4d9089f8250184f9b936710d912dR702).

Linking.md
```diff
@@ -524,10 +524,16 @@ segments][passive_segments] to prevent each thread from reinitializing
...
 reinitialized every time the module is instantiated on a new WebWorker as part
 of spawning a new thread. The `memory.init` instructions that initialize these
-passive segments will be emitted into a synthetic function, `__wasm_init_memory`
-that is called from `__wasm_call_ctors` before any constructors run. To make
-these `memory.init` instructions valid, a [DataCount section][datacount_section]
-will also be emitted.
+passive segments and the `data.drop` instructions that mark them collectible
+will be emitted into a synthetic function, `__wasm_init_memory` that is made the
+WebAssembly start function and called automatically on
+instantiation. `__wasm_init_memory` shall perform any synchronization necessary
```
sunfishcode: Just to be clear, this symbol name doesn't need to be exported, right? The function is just referenced by index in the start section.

Linking.md
aheejin: (Talked offline) I was confused because this means 'browser main thread' but I thought it was the main thread that is to be determined by the synchchronization routine in `__wasm_init_memory`. Changing it to 'browser main thread' might be little clearer.

Linking.md
tlively: Correct, I will clarify.

PR #131, opened 2019-10-18
Document __wasm_call_ctors

tlively:
A user rolling their own toolchain was asking after their missing
constructors on llvm-dev, and I was going to point them to the
documentation of __wasm_call_ctors wehen I discovered that there was
none. This PR fixes that hole in our docs.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -531,7 +531,9 @@ Start Section
 By default the static linker should not output a WebAssembly start
 section. Constructors are instead called from a synthetic function
 `__wasm_call_ctors` that the runtime and embedder should arrange to have called
-after instantiation.
+after instantiation. `__wasm_call_ctors` is not exported by default because it
+may be called by some other startup function defined by the runtime. For the
+embedder to call it directly it should be exported like any other function.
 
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -527,7 +528,10 @@ event symbols in the linked output.)
 Start Section
 -------------
 
-By default the static linker should not output a WebAssembly start section.
+By default the static linker should not output a WebAssembly start
+section. Constructors are instead called from a synthetic function
+`__wasm_call_ctors` that should be exported from the module and called after
+instantiation by the WebAssembly runtime.
```
sbc100: Actually it just needs to called from somwhere.  For example in WASI (and now sometimes in emscripten too) its gets called from `_start` which is linked in as `crt1.o`.

sbc100: For this reason it isn't actually exported by default by wasm-ld.  If you want to call it from the outside you need to `--export` it like any other function.

tlively: Ah, I guess `_start` then has to be exported and called by the runtime. I'll generalize the wording here and maybe change "runtime" to "embedder."

PR #134, opened 2019-11-25
Rename argc/argv `main` to `__main_argc_argv`

sunfishcode:
Change the ABI for the user entry point to rename `main` to
`__main_argc_argv` when it has argc/argv. This is needed because
wasm requires caller and callee signatures to match, so the
traditional trick of passing main arguments even if it doesn't
expect them doesn't work.

LLVM and related tools have been using the name `__original_main`
for a similar purpose, but that's a confusing name.

There's also a change here in that this is renaming the argc/argv
form rather than renaming the nullary form. The choice is somewhat
arbitrary, but I think it's slightly nicer to bias the aesthetics
toward the no-argument form, because that's the smaller and
simpler form.

### DIFF
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -154,3 +154,26 @@ is no way to traverse the linear stack. There is also no currently-specified way
 is used as the frame pointer or base pointer. This functionality is not needed for backtracing or unwinding (since the
 wasm VM must do this in any case); however it may still be desirable to allow this functionality for debugging
 or in-field crash reporting. Future ABIs may designate a convention for determining frame size and local usage.
+
+## Program startup
+
+### User entrypoint
+
+In the wasm C ABI, `main` is used for a user entrypoint which takes no
+arguments, eg.  `main(void)` or `main()` in C or C++. On other platforms,
+it's common to also pass `main` the `argc` and `argv` arguments, but on
+wasm this trick is complicated by the fact that wasm requires callers and
+callees to have matching signatures. So, in wasm, `__main_argc_argv` is
+used for a user entrypoint which takes `argc` and `argv` parameters, eg.
+`main(int argc, char *argv[])`.
+
+In the wasm C ABI, there is no `envp` parameter to the user entrypoint. Since
+wasm's signature rule makes this awkward to support, and since it's not required
+by C, POSIX, or any other relavant standard, and since it's generally considered
+obsolete in favor of `getenv` anyway, it's not supported in the wasm C ABI.
+
+### Program entrypoint
+
+The program entrypoint (traditionally named `_start` on other platforms)
+isn't considered part of the C ABI per se, and may differ depending on
+what environment the program will be run in.
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -169,7 +169,7 @@ used for a user entrypoint which takes `argc` and `argv` parameters, eg.
 
 In the wasm C ABI, there is no `envp` parameter to the user entrypoint. Since
 wasm's signature rule makes this awkward to support, and since it's not required
-by C, POSIX, or any other relavant standard, and since it's generally considered
+by C, POSIX, or any other relevant standard, and since it's generally considered
 obsolete in favor of `getenv` anyway, it's not supported in the wasm C ABI.
 
 ### Program entrypoint
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -159,21 +159,41 @@ or in-field crash reporting. Future ABIs may designate a convention for determin
 
 ### User entrypoint
 
-In the wasm C ABI, `main` is used for a user entrypoint which takes no
-arguments, eg.  `main(void)` or `main()` in C or C++. On other platforms,
-it's common to also pass `main` the `argc` and `argv` arguments, but on
-wasm this trick is complicated by the fact that wasm requires callers and
-callees to have matching signatures. So, in wasm, `__main_argc_argv` is
-used for a user entrypoint which takes `argc` and `argv` parameters, eg.
-`main(int argc, char *argv[])`.
+The *user entrypoint* is the function which runs the bulk of the program.
+It is called `main` in C, C++, and other languages. Note that this may
+not be the first function in the program to be called, as programs may
+also have global constructors which run before it.
 
-In the wasm C ABI, there is no `envp` parameter to the user entrypoint. Since
-wasm's signature rule makes this awkward to support, and since it's not required
-by C, POSIX, or any other relevant standard, and since it's generally considered
-obsolete in favor of `getenv` anyway, it's not supported in the wasm C ABI.
+At the wasm C ABI level, the following symbol names are used:
+
+C ABI Symbol name            | C and C++ signature                |
+---------------------------- | -----------------------------------|
+`main`                       | `int main(void)` or `int main()`   |
+`__main_argc_argv`           | `int main(int argc, char *argv[])` |
+
+These symbol names only apply at the ABI level; C and C++ source should
+continue to use the standard `main` name, and compilers will handle the
+details of conforming to the ABI.
+
+Also note that C symbol names are distinct from WebAssembly export
+names, which are outside the scope of the C ABI. Toolchains which export
+the user entrypoint may chose to export it as the name `main`, even when
+the C ABI symbol name is `__main_argc_argv`.
+
+A symbol name other than `main` is needed because the usual trick of
+having implementations pass arguments to `main` even when they aren't
+needed doesn't work in wasm, which requires caller and callee signatures
+to exactly match.
+
+For the same reason, the wasm C ABI doesn't support an `envp` parameter.
+Fortunately, `envp` is not required by C, POSIX, or any other relevant
+standards, and is generally considered obsolete in favor of `getenv`.
 
 ### Program entrypoint
 
-The program entrypoint (traditionally named `_start` on other platforms)
-isn't considered part of the C ABI per se, and may differ depending on
-what environment the program will be run in.
+The *program entrypoint* is the first function in the program to be called.
+It is commonly called `_start` on other platforms, though this is a
+low-level detail that most code doesn't interact with.
+
+The program entrypoint is out of scope for the wasm C ABI. It may depend
+on what environment the program will be run in.


sunfishcode:
As a heads-up, as I was writing this up, it occurred to me that renaming the argc/argv form would be a little nicer than renaming the no-arg form, but I forgot to adjust the title before submitting the pull request. I've now corrected the title.

BasicCABI.md
```diff
@@ -154,3 +154,26 @@ is no way to traverse the linear stack. There is also no currently-specified way
...
+arguments, eg.  `main(void)` or `main()` in C or C++. On other platforms,
+it's common to also pass `main` the `argc` and `argv` arguments, but on
+wasm this trick is complicated by the fact that wasm requires callers and
+callees to have matching signatures. So, in wasm, `__main_argc_argv` is
+used for a user entrypoint which takes `argc` and `argv` parameters, eg.
+`main(int argc, char *argv[])`.
+
+In the wasm C ABI, there is no `envp` parameter to the user entrypoint. Since
+wasm's signature rule makes this awkward to support, and since it's not required
+by C, POSIX, or any other relavant standard, and since it's generally considered
```
kripken: typo in `relevant`

sunfishcode:
Yes, that's right. It simplifies things to say that we just rename `main`, rather than that we keep `main` around and add a wrapper.

I also plan to submit a clang patch to do this rename in the frontend, instead of in the LLVM backend, to eliminate the wrapper, which has turned out to be a nuisance in practice.

kripken:
In that case I'm not that happy about the part where we export a different name from the wasm module. That's more complicated than the current single export, and is user-visible and perhaps confusing.

sunfishcode:
I've now posted a clang patch which implements the `__main_argc_argv` proposal [here](https://reviews.llvm.org/D70700).

It's a name mangling rule, which means:
 - If users call `main` from other translation units, those callsites are also renamed to match.
 - When debug info is emitted, it uses the user's name "main".

It's very simple. There's no generated wrapper function, linking works the same way for LTO and non-LTO, and libc code can easily detect whether to link in the command-line argument initialization code by defining its own `main` function in libc.a:

```c
__attribute__((weak))
int main(void) {
    int argc = ...;
    char **argv = ...;
    return __main_argc_argv(argc, argv);
}
```
It also isn't that different from the `__original_main` situation we have right now, where Binaryen is special-case inlining it. If you inline `__main_argc_argv` into `main` here, you'll end up with a function called `main` containing the user main code, so backtraces etc. will look like you want them to.

kripken:
Sorry @sunfishcode , I'm not sure for your answer whether or not it addresses the topic of a different export name?

sunfishcode:
It doesn't change the `_start` export. I'd ideally like to rename `_start` also, but that's a bigger change we can talk about separately.

kripken:
I apologize, maybe I wasn't clear before. I wasn't talking about `_start` which I agree is a separate issue. What I'm asking about is what I asked about earlier too, that this proposal makes the wasm module export either `main` or `__main_argc_argv`. I wasn't sure if your last comment refers to this (as you mention other callsites etc. which I think may involve exports?).

sunfishcode:
Ah; in wasi-sdk, `main` isn't exported from the final wasm binary. I guess in Emscripten it is exported, or it can be?

It is possible for a symbol name to differ from the wasm export name -- @sbc100's [export attribute patch](https://reviews.llvm.org/D70520) allows this, for example. So one possible option is to export `__main_argc_argv` as `main` at the wasm module level.

kripken:
Yes, in non-standalone wasm, emscripten exports `main` from the wasm binary. I wouldn't want to have a convention that where the export name can differ, so yeah, keeping the same export name as always sounds best.

I'm not sure what "entrypoint" means, in the text of this PR, if it isn't an export? Did we define it somewhere?

sunfishcode:
I've now rewritten the wording to give a definition of "entrypoint" and mention that export names may differ from ABI symbol names, and to hopefully be more clear overall. Let me know what you think!

sbc100:
Right now emscripten has two entry-point modes.  The primary mode is where "main" is exported and called directly from JS after the constructors have been called.    In this mode there is no `_start`, only `main`.

The second mode we what we call "standalone wasm mode" where `_start` is exported and main is not.

I'm not sure how this change will effect the first/primary mode that emscripten uses, but in general it looks like a good change.   We currently export main by passing `--export=main` which is a linker flag (and which will work with wask-sdk too btw).    I think this change as proposed might require some emscripten-side changes.   Since we assume we can pass args to `main` from JS and due to the JS calling conventions that works for 0-arg and 2-arg main both.

If we can move emscripten over to always use `_start` then this problem would go away.  Do you know of any reasons why that might be a bad idea @kripken ?

kripken:
@sbc100 Emscripten changing from `main` to `_start` for the export to be called might be disruptive for existing users, that call main themselves directly. (Also, I do think `main` is the better name!) So I'd prefer to not do that, but it looks like this PR isn't?

The current wording looks mostly reasonable to me, but I don't know the linking process well enough to say if there is any downside to this over the current `__original_main`. @sbc100 do you think it would have problems with our standalone support in emscripten?

sbc100:
To support the status quo in emscripten we would need way to export a symbol called `main` which takes args.    I sounds like under this proposal `--export=main` would result in the zero arg main being exported.

sbc100:
I'm not a fan of the current `__original_main` mangling or the way its found its way into our crt1.c code in both wasi and emscripten, so if we can move away from that I would be happy.

sunfishcode:
As a possible path forward, I've now updated the clang patch which implements this: https://reviews.llvm.org/D70700 
so that it checks the target, and doesn't do the mangling on Emscripten. That way, Emscripten is unaffected by this change.

sunfishcode:
@sbc100 @kripken Any update here? I need some solution to this problem in order to enable LTO builds of libc, and right now, compatibility with Emscripten appears to be the only thing holding this back.

kripken:
I don't quite understand what that patch does (not familiar with that part of clang), but if it doesn't affect emscripten then I have no objection to it myself. (That seems separate from the discussion in this PR, but I don't think we have a clear answer here?)

BasicCABI.md
```diff
@@ -154,3 +154,46 @@ is no way to traverse the linear stack. There is also no currently-specified way
...
+`__main_argc_argv`           | `int main(int argc, char *argv[])` |
+
+These symbol names only apply at the ABI level; C and C++ source should
+continue to use the standard `main` name, and compilers will handle the
+details of conforming to the ABI.
+
+Also note that C symbol names are distinct from WebAssembly export
+names, which are outside the scope of the C ABI. Toolchains which export
+the user entrypoint may chose to export it as the name `main`, even when
+the C ABI symbol name is `__main_argc_argv`.
```
sbc100: But can the toolchain support exporting main?   Won't this a toolchain that want to do something like `--export=main`?   We need to some way to export the user entry point and it seems like with this change it because hard to name the user entry point because it has two possible names, no?

sunfishcode: I can confirm that `__attribute__((export_name("main")))` does work with this patch, with and without argv/argc. I get an export named "main" regardless of the internal symbol name.

However, `-Wl,--export=main` indeed doesn't work with this patch in the case of `main` with argc/argv. One way to fix that would be to introduce a -mexport=main flag to clang and have people use that instead. Clang could translate that into inserting the appropriate `export_name` directive. Clang flags are nicer for end users than `-Wl,` flags anyway, and it would handle mangled names better in general. What do you think?

sbc100: For emscripten can't ask users to decorate the main symbol in the source code.

A new `-mexport=main` cflag would have to be applied to every single compilation unit.    Seems kind of heavy weight and would require more plumbing to land in llvm.

Actually I think may be able to just pass `-Wl,--export=main` and `-Wl,--export=__main_argc_argv` since this flag doesn't warn or error if the symbol is missing, and then we can have binaryen take care of exporting __main_argc_argv as main perhaps? Will have to see how it goes.

BasicCABI.md
```diff
@@ -154,3 +154,46 @@ is no way to traverse the linear stack. There is also no currently-specified way
...
+standards, and is generally considered obsolete in favor of `getenv`.
+
+### Program entrypoint
+
+The *program entrypoint* is the first function in the program to be called.
+It is commonly called `_start` on other platforms, though this is a
+low-level detail that most code doesn't interact with.
+
+The program entrypoint is out of scope for the wasm C ABI. It may depend
+on what environment the program will be run in.
```
sbc100: Is it out of scope?   Were do we document `_start` and `__wasm_call_ctors` if not here?

sunfishcode: `__wasm_call_ctors` is a convention between the linker and crt1.o (or something else serving the purpose of crt1.o), making it appropriate for Linking.md, where it is. Anyone calling `__wasm_call_ctors` has to know that the module was generated from a linker following Linking.md.

I see `_start` as being a convention between a wasm module and the outside which doesn't know anything about the tools that produced it. It may not use BasicCABI.md or Linking.md. Eventually it'll be replaced by something we define in [interface types](https://github.com/WebAssembly/interface-types/blob/linking/proposals/interface-types/linking/Explainer.md), but for now it's just a simple "_start" convention.

sunfishcode:
In the LLVM patch, this change is disabled when the target triple OS is Emscripten. Also, @sbc100's comments [above](https://github.com/WebAssembly/tool-conventions/pull/134#discussion_r360550111) suggest there might be a reasonable path for implementing this in Emscripten when it's ready as well. With that, are there any remaining concerns about landing this and proceeding with the [changes to LLVM](https://reviews.llvm.org/D70700), [wasi-libc](https://github.com/CraneStation/wasi-libc/pull/152)?

sbc100:
I support this change, but I'd be sad if can't find a way to make this work in emscripten too.  I fairly certain that we can but I've not invested any time in it yet.   I can you give me a little more time to prototype.   Maybe by EOD Wednesday?  If don't find any fundamental issues by then we can go ahead and land?

sunfishcode:
@sbc100 Sounds good to me.

sunfishcode:
It's now EOD on Wednesday; is there an update?

sbc100:
Did you verify that the debug name in the dwarf section is correct?  I couldn't see the functions names in the output of `llvm-dwarfdump --all` which means I'm probably doing something wrong.

I did confirm that the name section does have the mangled name in it, despite then name section supposedly containing the demangled names.   This I believe is because the name section is generated by the linker by demanging the mangled names and I guess doesn't (yet) know how to demangle `__main_argv_argv` back into `main`.

sunfishcode:
The llvm-dwarfdump --all output contains the name "main" as the `DW_AT_name` field for me (and "__main_argc_argv" in the DW_AT_linkage_name, as expected).

Confirmed about the name section; I propose the following patch, which fixes it:
```diff
diff --git a/lld/wasm/Symbols.cpp b/lld/wasm/Symbols.cpp
index 88e6f217c0d..0d4f7f648d5 100644
--- a/lld/wasm/Symbols.cpp
+++ b/lld/wasm/Symbols.cpp
@@ -29,6 +29,8 @@ std::string toString(const wasm::Symbol &sym) {
 }

 std::string maybeDemangleSymbol(StringRef name) {
+  if (name == "__main_argc_argv")
+    return "main";
   if (wasm::config->demangle)
     return demangleItanium(name);
   return std::string(name);
```

sbc100:
OK, I've played with the patch in emscripten and I think we have a way forward.  I'm going ahead with this and the llvm patch. 

I have provisional patch for binaryen and emscripten so we should be able remove the emscripten exclusion from the llvm patch (either before or after landing it).

PR #146, opened 2020-08-03
R_WEBASSEMBLY_* => R_WASM_*

kripken:
I think this is just a typo or left over from some old
convention?


sbc100:
Doh!

sbc100:
Search and replace fail.

PR #147, opened 2020-09-24
Update file to reflect current defined relocations in ecosystem

pmatos:
Include information about clang/LLVM versions where each relocation is implemented.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -2,11 +2,11 @@ WebAssembly Object File Linking
 ===============================
 
 This document describes the WebAssembly object file format and the ABI used for
-statically linking them to produce an executable WebAssembly module.  This is
-currently implemented in the clang/LLVM WebAssembly backend (as of LLVM 8.0) and
-other tools such as binaryen and wabt.  As mentioned in [README](README.md),
-this is not part of the official WebAssembly specification and other runtimes
-may choose to follow a different set of linking conventions.
+statically linking them to produce an executable WebAssembly module. This is
+currently implemented in the clang/LLVM WebAssembly 
+backend and other tools such as binaryen and wabt.  As mentioned in 
+[README](README.md), this is not part of the official WebAssembly specification 
+and other runtimes may choose to follow a different set of linking conventions.
 
 Overview
 --------
@@ -35,10 +35,10 @@ to disassemble the code section.  The extra metadata required by the linker
 is stored in a custom ["linking"](#linking-metadata-section) section and zero or
 more relocation sections whose names begin with "reloc.".  For each section that
 requires relocation a "reloc" section will be present in the wasm file.  By
-convention the reloc section names end with the name of the section that they refer
-to: e.g. "reloc.CODE" for code section relocations.  However everything after
-the period is ignored and the specific target section is encoded in the reloc
-section itself.
+convention the reloc section names end with the name of the section that they 
+refer to: e.g. "reloc.CODE" for code section relocations.  However, everything 
+after the period is ignored and the specific target section is encoded in the 
+reloc section itself.
 
 The linker additionally checks that linked object files were built targeting
 compatible feature sets. Unlike native targets, WebAssembly has no runtime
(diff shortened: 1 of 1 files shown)


pmatos:
Looking to get comments on the updates. I included some relocation 20 which has not been implemented in clang/LLVM yet but are working on it. So I included a note below the relocation list explaining which clang/LLVM versions contain which relocations.

wingo:
LGTM, for what it's worth.  Style nit, I would put suffixes like "(Since LLVM 10.x)" on the relocs added since LLVM 10.x (and 11.x, and presumably 12.x); otherwise with the summary below, it's easy to miss.

Linking.md
```diff
@@ -2,9 +2,8 @@ WebAssembly Object File Linking
 ===============================
 
 This document describes the WebAssembly object file format and the ABI used for
-statically linking them to produce an executable WebAssembly module.  This is
-currently implemented in the clang/LLVM WebAssembly backend (as of LLVM 8.0) and
-other tools such as binaryen and wabt.  As mentioned in [README](README.md),
+statically linking them to produce an executable WebAssembly module.  This and most of the relocations are
+currently implemented in the clang/LLVM WebAssembly backend and other tools such as binaryen and wabt.  As mentioned in [README](README.md),
```
sbc100, line 6: Can you wrap these lines at 80?

sbc100: I'm not sure why you have adding "and most of the relocations" here.

Linking.md
```diff
@@ -107,30 +106,48 @@ A relocation type can be one of the following:
 - `10 / R_WASM_EVENT_INDEX_LEB` - an event index encoded as a 5-byte
   [varuint32]. Used for the immediate argument of a `throw` and `if_except`
   instruction.
-- `13 / R_WASM_TABLE_NUMBER_LEB` - a table number encoded as a 5-byte
+- `11 / R_WASM_MEMORY_ADDR_REL_SLEB` - see documentation on [Dynamic Linking](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md).
+- `12 / R_WASM_TABLE_INDEX_REL_SLEB` - see documentation on [Dynamic Linking](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md).
```
sbc100, line 110: We deliberately do not including the dynamic linking relocations in this document because the dynamic linking stuff is not settled at this point.  Its considered to be an emscripten-specific stop-gap solution.

Linking.md
```diff
@@ -107,30 +106,48 @@ A relocation type can be one of the following:
...
+[varuint64]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varuintn
 [varint32]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varintn
+[varint64]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#varintn
 [uint32]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#uintn
+[uint64]: https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#uintn
 
 For `R_WASM_MEMORY_ADDR_LEB`, `R_WASM_MEMORY_ADDR_SLEB`,
 `R_WASM_MEMORY_ADDR_I32`, `R_WASM_FUNCTION_OFFSET_I32`, and
-`R_WASM_SECTION_OFFSET_I32` relocations the following field is additionally
+`R_WASM_SECTION_OFFSET_I32` relocations (and their 64-bit counterparts) the following field is additionally
```
sbc100, line 133: Wrap at 80 here and below.

sbc100 (commented):
I'm also not sure the 64-bit relocation types belong in this document yet since there are very new and experimental.   Perhaps we could separate out the 64-bit relocation stuff into its own document or perhaps its own section?  Or perhaps a footnote/warning is enough?

What do you think @aardappel @sunfishcode

sunfishcode (commented):
Including the 64-bit relocations in this document seems ok to me. +1 to adding warnings/footnotes that the 64-bit relocations aren't yet stable.

Linking.md
```diff
@@ -107,30 +106,48 @@ A relocation type can be one of the following:
 - `10 / R_WASM_EVENT_INDEX_LEB` - an event index encoded as a 5-byte
   [varuint32]. Used for the immediate argument of a `throw` and `if_except`
   instruction.
-- `13 / R_WASM_TABLE_NUMBER_LEB` - a table number encoded as a 5-byte
+- `11 / R_WASM_MEMORY_ADDR_REL_SLEB` - see documentation on [Dynamic Linking](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md).
+- `12 / R_WASM_TABLE_INDEX_REL_SLEB` - see documentation on [Dynamic Linking](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md).
+- `13 / R_WASM_GLOBAL_INDEX_I32` - a global index encoded as [uint32].
+- `14 / R_WASM_MEMORY_ADDR_LEB64` - the 64-bit counterpart of `R_WASM_MEMORY_ADDR_LEB`. A 64-bit linear memory index encoded as a 10-byte [varuint64], Used for the immediate argument of a `load` or `store` instruction on a 64-bit linear memory array.
```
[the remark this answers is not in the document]
sbc100: I think that makes sense.  The names themselves are changeable.  What do you think @wvo?

sbc100:
> Including the 64-bit relocations in this document seems ok to me. +1 to adding warnings/footnotes that the 64-bit relocations aren't yet stable.

I assume you would still prefer not to include the REL relocations used for PIC code?

pmatos:
Thanks for the reviews - I will prepare the necessary changes, commit then and re-request reviews.

Linking.md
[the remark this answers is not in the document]
aardappel: sgtm! would be happy to make that change if no-one else is.

aardappel (approved):
Thanks for updating this! I should have probably done this, though much like the dynamic linking the wasm64 stuff is not necessarily settled yet.. but there's unlikely to be big changes so I am fine adding it now.

sunfishcode:
> I assume you would still prefer not to include the REL relocations used for PIC code?

Yep. As @aardappel says, wasm64 seems unlikely to see big changes, and it's similar enough to wasm32 that keeping them together will make it easier to keep them in sync. But I'm expecting PIC and dynamic linking will see big changes in the future.

Linking.md
pmatos: Done - https://github.com/WebAssembly/tool-conventions/pull/147/commits/541c11b53ff1539b19c9946f4aac587c3d05f072

Linking.md
pmatos: I have removed those at https://github.com/WebAssembly/tool-conventions/pull/147/commits/d61709c4062d3b1a220430999fe4321909908966

Linking.md
pmatos: This is now complete.

pmatos:
Most of the changes have been completed, re-requesting reviews.

pmatos:
Regarding the renaming suggested by @sunfishcode , I can do them under this PR if needed but it will require changes to LLVM and wabt and therefore invalidate the versions from which the relocs were available, although I can mention old names as well.

How do you prefer to do this?

sbc100:
I think we can land this as is and do the rename a followup since it will effect all relocation names.

Linking.md
```diff
@@ -2,11 +2,11 @@ WebAssembly Object File Linking
 ===============================
 
 This document describes the WebAssembly object file format and the ABI used for
-statically linking them to produce an executable WebAssembly module.  This is
-currently implemented in the clang/LLVM WebAssembly backend (as of LLVM 8.0) and
-other tools such as binaryen and wabt.  As mentioned in [README](README.md),
-this is not part of the official WebAssembly specification and other runtimes
-may choose to follow a different set of linking conventions.
+statically linking them to produce an executable WebAssembly module.  These 
+relocations are currently implemented in the clang/LLVM WebAssembly
```
sbc100, line 6: Why this change?  I think "This" was referring to the entire ABI, of which the relocations I see as just part of.   Am I wrong?

Linking.md
```diff
@@ -583,7 +615,7 @@ Thread Local Storage
 --------------------
 
 Currently, thread-local storage is only supported in the main WASM module
-and cannot be accessed outside of it. This corresponds to the ELF local
+and cannot be accessed outside it. This corresponds to the ELF local
```
sbc100, line 618: Hmm .. reading in sentence I realize I don't really no what this mean.   Is `main` refering to emscripten's MAIN_MODULE?    If so that terminology does not make sense in this document.  Anyway, that is unrelated to this PR.   

I'd be tempted just to revert this change in the interest of keep this PR narrow.   Do you find " "outside it" to be more grammatically correct that "outside of it"?   They both some reasonable to me.

sbc100 (approved):
lgtm % a couple of nits

pmatos:
> I think we can land this as is and do the rename a followup since it will effect all relocation names.

Sure!

Linking.md
pmatos: Well - to be fair it sounded odd, so I ran a grammar checker (langtool) and it suggested to change it so I did. I can revert it and we can look at it under another PR. Unsure what `main` is referring to but I definitely thought it was the emscripten `MAIN_MODULE` as I don't what else it could be about.

Linking.md
pmatos: Your right here - I made the change here initially with the narrow view of this as a description of the relocations. Will fix.

pmatos:
I have squashed the changes. If all is OK, feel free to merge.

PR #159, opened 2021-01-02
BasicCABI: LP64 means 32 bit int

douglasbagnall:
This matches the understanding found widely on the web, and the logic of the acronym.

It is also handy to have a type you know will be i32.


sunfishcode:
Yes, this is what LP64 means, and matches what's implemented in the `wasm64` target in development in clang.

aardappel:
Hey, it's still not too late to adopt LLP64 if we wanted :P

PR #164, opened 2021-04-13
Update linking docs

mwilliamson:
I've just finished updating a toy compiler to output object files instead of using the text format, so I thought it'd be good to include things that weren't clear to me from reading through the linking docs.

There are some details that I think would still be useful to include somewhere, although I'm not sure what the most appropriate place would be, nor am I sure whether there are alternative methods for the below. My impression, mostly from writing some C code and inspecting the resulting object file, is:

* rather than defining memory in the object file, memory should be imported from `env.__linear_memory`
* rather than defining a function table, one should be imported from `env.__indirect_function_table`
* on recent (>= 12) versions of LLVM, if `__wasm_call_ctors` is not called, then each exported function is wrapped in a function that first calls `__wasm_call_ctors` before calling the original function

I should also add that the fact that I was able to get things working within roughly a weekend, despite not being well versed in this sort of stuff, is a positive sign to some combination of an understandable design and good docs. So, thanks for putting this together in the first place!

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -236,6 +236,8 @@ subsection:
 | count       | `varuint32`  | number of init functions that follow  |
 | functions   | `varuint32*` | sequence of symbol indices            |
 
+The `WASM_INIT_FUNC` subsection must come after the `WASM_SYMBOL_TABLE` subsection.
+
 For `WASM_SYMBOL_TABLE` the following fields are present in the
 subsection:
 
--- a/Linking.md
+++ b/Linking.md
@@ -113,7 +113,7 @@ a 5-byte [varuint32], e.g. the type immediate in a `call_indirect`.
 - `8 / R_WASM_FUNCTION_OFFSET_I32` (since LLVM 10.0) - a byte offset within
 code section for the specific function encoded as a [uint32]. The offsets start
 at the actual function code excluding its size field.
-- `9 / R_WASM_SECTION_OFFSET_I32` (since LLVM 10.0) - an byte offset from start
+- `9 / R_WASM_SECTION_OFFSET_I32` (since LLVM 10.0) - a byte offset from start
 of the specified section encoded as a [uint32].
 - `10 / R_WASM_EVENT_INDEX_LEB` (since LLVM 10.0) - an event index encoded as a
 5-byte [varuint32]. Used for the immediate argument of a `throw` and `if_except`
--- a/Linking.md
+++ b/Linking.md
@@ -106,7 +106,7 @@ instruction, e.g. taking the address of a C++ global.
 - `5 / R_WASM_MEMORY_ADDR_I32` (since LLVM 10.0) - a linear memory index
 encoded  as a [uint32], e.g. taking the address of a C++ global in a static data
 initializer.
-- `6 / R_WASM_TYPE_INDEX_LEB` (since LLVM 10.0) - a type table index encoded as
+- `6 / R_WASM_TYPE_INDEX_LEB` (since LLVM 10.0) - a type index encoded as
 a 5-byte [varuint32], e.g. the type immediate in a `call_indirect`.
 - `7 / R_WASM_GLOBAL_INDEX_LEB` (since LLVM 10.0) - a global index encoded as a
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -57,10 +57,19 @@ A relocation section is a user-defined section with a name starting with
 section they apply to, and must be sequenced in the module after that
 section.
 
+Relocation sections can only target code, data and custom sections.
+All other sections are synthetic sections: that is, rather than being
+`memcpy`'d into place as the code and data sections are, they
+are created from scratch by the linker.
```
sbc100, line 63: Looks like this paragraph could use re-flowing to 80 char lines.

Linking.md
```diff
@@ -57,10 +57,19 @@ A relocation section is a user-defined section with a name starting with
...
+
 The "reloc." custom sections must come after the
 ["linking"](#linking-metadata-section) custom section in order to validate
 relocation indices.
 
+Any LEB128-encoded values should be maximally padded so that they can be rewritten
+without affecting the position of any other bytes. For instance, the function index
+3 should be encoded as `0x83 0x80 0x80 0x80 0x00`.
```
sbc100, line 71: We do specify this already for each relocation type. e.g. 

```
- `0 / R_WASM_FUNCTION_INDEX_LEB` - a function index encoded as a 5-byte         
  [varuint32]. Used for the immediate argument of a `call` instruction.   
```

But I think its find to re-iterate.

sbc100 (commented):
Nice!   Thanks for working on this.

Linking.md
mwilliamson: Sure, done!

Linking.md
mwilliamson: Yeah, I thought about whether it was redundant, but I thought being explicit about the general case with reasoning and an example might be helpful.

mwilliamson:
> There are some details that I think would still be useful to include somewhere, although I'm not sure what the most appropriate place would be, nor am I sure whether there are alternative methods for the below. My impression, mostly from writing some C code and inspecting the resulting object file, is:
> 
>     * rather than defining memory in the object file, memory should be imported from `env.__linear_memory`
> 
>     * rather than defining a function table, one should be imported from `env.__indirect_function_table`
> 
>     * on recent (>= 12) versions of LLVM, if `__wasm_call_ctors` is not called, then each exported function is wrapped in a function that first calls `__wasm_call_ctors` before calling the original function

Did you have any thoughts on whether the linking docs are the right place to put these sorts of details?

sbc100:
> > There are some details that I think would still be useful to include somewhere, although I'm not sure what the most appropriate place would be, nor am I sure whether there are alternative methods for the below. My impression, mostly from writing some C code and inspecting the resulting object file, is:
> > ```
> > * rather than defining memory in the object file, memory should be imported from `env.__linear_memory`
> > 
> > * rather than defining a function table, one should be imported from `env.__indirect_function_table`
> > 
> > * on recent (>= 12) versions of LLVM, if `__wasm_call_ctors` is not called, then each exported function is wrapped in a function that first calls `__wasm_call_ctors` before calling the original function
> > ```
> 
> Did you have any thoughts on whether the linking docs are the right place to put these sorts of details?

I think that kind of stuff is more about the runtime ABI (i.e. that ABI used after everything is linked).  

That probably belongs more in either https://github.com/WebAssembly/tool-conventions/blob/master/BasicCABI.md or https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md

sunfishcode:
@sbc100 I was under the impression that `env.__linear_memory` etc. were .o file things here, since neither the BasicCABI nor WASI intend to require modules to import their linear memory.

sbc100:
Sorry yes, if you are talking about object files then we can/should mention `__linear_memory` and `__indirect_function_table` although I'm not actually sure how much the linker cares about those existing in object file.   I think they are mostly there to make the object files validate.    So I'd need to check how much we specify there..   at least for memory I'm pretty sure the linker doesn't care if the memory is imported or exports or what its name is.

PR #166, opened 2021-08-19
Note that the C ABI for vector types is not stable

tlively:
As suggested in https://reviews.llvm.org/D107502, which changes the ABI.


dschuff (approved):
this LGTM from the ABI perspective.

I'm also thinking about the practical consequence for toolchain users. For them, the stable ABI mostly affects calling convention compatibility and sharing data structures in memory across time/processes/parties. I guess most users aren't writing public data structures with vector types, or public functions which take vector types as arguments, so changes aren't likely to break them.

tlively:
Right, and if I understand correctly, LLVM does have a stable in-memory representation for vectors, so the common case of passing pointers to vector types should be stable at least. I don't know if we should say anything about that in this document, though.

Issue #179, opened 2022-02-02
Rename Code Annotations to Code Metadata


yuri91:
In the process of rebasing the [Branch Hinting](https://github.com/WebAssembly/branch-hinting) proposal on top of the [Custom Annotations](https://github.com/WebAssembly/annotations) proposal,
I realized that the name "Code Annotations" is too confusing.

The term "annotation" is already used by the Custom Annotations proposal to refer to
an analogous of sections in text format.

For example, the proposal adds a "Name annotation" corresponding to the "name section".

Adding Code Annotations to the mix makes it very confusing: A "Code Annotation section"
would have its textual analogous named "Code Annotation annotation".

I propose to rename "Code Annotations" to "Code Metadata".

I feel it is a more descriptive name in itself, and it allows to have "Code Metadata sections"
and "Code Metadata annotations" without any confusion.

We could consider just "Metadata", but I feel that it may be a too general term that could prevent
other kinds of metadata being named as such in the future.

The section name in binary format would turn from `code_annotation.<feature>`
to something like `meta.code.<feature>`.
An alternative could be `code.meta.<feature>` or just `meta.<feature>` but I think it is nicer if the "meta" acts
as a prefix for possible future kinds of metadata.

In the text format `@code_annotation.<feature>` would become `@meta.code.<feature>`.

I am interested in getting feedback about this, in particular from people that
are currently experimenting with custom annotations.

As far as I know, other than the Branch Hinting proposal, only the [Instrument and Tracing Technology](https://github.com/WebAssembly/instrument-tracing)
proposal uses custom annotations (ping @jacob-abraham and @rrwinterton), and @tlively  mentioned in a CG meeting
that he wanted to experiment with a non-standard custom annotation in llvm.

tlively:
Disambiguating names sounds reasonable to me. IMO `code.meta.<feature>` or `meta.code.<feature>` does not seem as descriptive as `code_annotation.<feature>`, though. Maybe it doesn't matter because `<feature>` will always be descriptive enough.

yuri91:
Do you think that also the name "Code Metadata" for the feature itself is less descriptive, or just the section/annotation strings?

Since the goal of this change would be to make the spec/explanation less confusing, I  don't want to end up making thing worse on that front.

tlively:
I think "code metadata" is a good name. It even suggests to me that it is not supposed to have semantically observable affects on the code, which was less clear with "code annotations."

[...]

PR #180, opened 2022-02-07
Rename Code Annotations to Code Metadata

yuri91:
The previous name was too similar to the "Custom Annotations" proposal.

See #179 for the full context.


CodeMetadata.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+Where `funcpos` is the byte offset of the annotation starting from the beginning of the function body,  and `data` is a further payload, whose content depends on the section type `A`.
+
+`codemetadatafunc` entries must appear in order of increasing `idx`, and duplicate `idx` values are not allowed.
+`codemetadatainstance` entries must appear in order of increasing `funcpos`, and duplicate `funcpos` values are not allowed.
+
+## Text Representation
+
+Code Metadata are representend in the .wat format using [custom annotations](https://github.com/WebAssembly/annotations), as follows:
```
tlively, line 48:
```suggestion
Code Metadata are represented in the .wat format using [custom annotations](https://github.com/WebAssembly/annotations), as follows:
```

tlively (approved):

CodeMetadata.md
yuri91: thanks!

yuri91:
I am about to propagate this change to other projects (The branch hinting proposal, v8's branch hinting implementation,  wabt implementation PR, Cheerp's branch hinting support).

If there are no issues it would be nice to have it merged so I can more easily refer to the new terminology in these other projects.

Issue #193, opened 2022-12-03
Mention elemidx and dataidx in the Linking.md documentation


matovitch:
Hello,

I am writing a wasm backend for my compiler and I was planning to output relocatable wasm modules.

In the [Linking.md](https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md) document, there is no mention of `dataidx` and `elemidx` renumbering such that we would need to pad the LEB128 encodings of these to 5 bytes. I believe this should be new relocation types but these do not currently exist because LLVM does not support the element section or generate any `memory.init` or `data.drop` instruction.

I lost quite a bit of time wondering why the documentation did not address this issue. So here are my questions:

First, did I understood the problem properly and there is no mention of data or element segments reordering because these are not generated/supported by LLVM/lld?
Second, if this is correct, can I submit a PR updating the markdown documentation?

Thank you for your help. :)

sbc100:
Yes, llvm doesn't have any support for generating first class data or element segments, so there are no relocation types for these, and no support in the linker.

If you would like to support such things a good place to start would be looking at adding support to lld (wasm-ld), and most likely the llvm assembly format (.s format).

BTW, is the existing support for data symbols not sufficient for your language?    Do you really need first class data segments and element segments?  I'd like to understand your use case here.

matovitch:
No it is probably sufficient for my language (it is sufficient for C++ after all ^^). I implemented the complete basic spec to generate non relocatable modules and was wondering how to adapt my code for instructions like `data.drop`, `elem.drop`, `memory.init`, `table.init`...etc. I can just avoid generating these instructions in the first place. But a mention in the documentation explaining why these types do not exist would have save me a bit of time wondering why these were missing.

sbc100:
Yes, there is currently no support for data.drop, elem.drop, memory.init, table.init, etc in the linkable format.  We should add a "limitations" section I guess.

PR #194, opened 2022-12-03
Add limitations section to Linking.md to clarify element and data segment indices renumbering is not supported (#193)

matovitch:
See issue #193.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -24,7 +24,7 @@ tasks need to be performed:
 - Merging of globals sections (re-numbering globals)
 - Merging of event sections (re-numbering events)
 - Merging of table sections (re-numbering tables)
-- Merging of data segments (re-positioning data)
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -686,3 +686,12 @@ have its address computed as follows:
 
 The variable can then be used as normal. Upon thread exit, the runtime should free
 the memory allocated for the TLS block.
+
+### Limitations
+
+Note data segments are only re-positioned, i.e. only the offsets of active data
```
sbc100, line 692: How about rephasing this as.


```
- Currently only active data segments are supported.  The relocation types necessary for referencing passive segment (e.g. in `data.drop` or `memory.init` instruction) do not yet exist.
- There is currently no support for table element segments, either active or passive.
```


```

matovitch: I like your reformulation, clear and to the point. I've updated the PR.

PR #201, opened 2023-02-22
Add JavaScript and Javy as known names in producers

jeffcharles:
[Javy](https://github.com/Shopify/javy) is a tool that takes JavaScript source code as input and outputs a WebAssembly module that executes that code.


dschuff (approved):
Thanks!

PR #207, opened 2023-04-28
`Coredump.md`: Add support for multiple instances in a coredump

### LINKED ISSUE #204: Coredumps and multiple wasm modules
How are we supposed to represent coredumps when there are multiple modules involved? For example, these two modules are linked together and so `B.g` calls `A.f` which traps:

```wat
(module $A
  (memory ...)
  (func (export "f")
    unreachable))

(module $B
  (import "" "f" (func))
  (memory ...)
  (func (export "g")
    call 0))
```

It is easy to create such examples via JS doing the linking on the Web or with (for example) `wasmtime::Linker` elsewhere.

It seems like we need a module name or something in each stack frame. But actually module name isn't precise enough because the same module could be instantiated multiple times. We need a way to identify instances in the store.

If each module has its own memory, we can encode that in the current format, assuming that the coredump Wasm module relying on multi-memory is acceptable. However there is still the question of how to map the coredump's memories back to the original instances. This also applies to globals.[^tables]

[^tables]: I guess coredumps capturing tables is a non-goal since they aren't mentioned in the coredumps document? Definitely trickier. Things we could do, but I won't discuss them here. If interested I can open a new issue.

To address these issues, I think that there should be additional coredump sections to establish module and instance index spaces:

```
coredump-modules ::= customsec(vec(coredump-module))
coredump-module ::= 0x0 module-name:name
                  | 0x1 module-bytes:module

coredump-instances ::= customsec(vec(coredump-instance))
coredump-instance ::= 0x0 module:u32 memories:vec(u32) globals:vec(u32)
```

Here modules are either defined by name/URL/path or I also gave the option of bundling the whole Wasm module inline, which could be useful to avoid wrangling paths/URLs/etc in some scenarios. But if we don't want that second variant, we could just have names. Just throwing it out there.

Each `coredump-instance` describes what module it is an instance of (via index into the coredump modules index space), as well as which memories in the coredump Wasm file are its memories and which globals in the coredump Wasm file are its globals. That is, memory `i` in a `coredump-instance` is `memories[coredump-instance.memories[i]]` from the coredump wasm file's memory section, and similar for globals.

And finally, `frame` productions inside the `"corestack"` section would grow an `instanceidx:u32` identifier:

```
frame       ::= 0x0 instanceidx:u32 funcidx:u32 codeoffset:u32
                locals:vec(value) stack:vec(value)
```

Thoughts?

cc @xtuc @itsrainy @dschuff @alexcrichton

fitzgen:
@dschuff PTAL

cc @xtuc 

One thing that I noticed is that it seems like the `core` section isn't really useful anymore: all it does is define an `executable-name`, which I think was intended to be the module name before, but now we support multiple modules. I left it as-is in this PR but we should discuss whether we want to remove it, change it, or something else.

Fixes #204

### DIFF
--- a/Coredump.md
+++ b/Coredump.md
@@ -21,6 +21,7 @@ Stability of this specification is **experimental**.
 
 Tools that support the generation of Wasm coredumps:
 - [wasm-edit]
+- [Wasmtime]
 
 Debugger that supports post-mortem debugging with Wasm coredumps:
 - [wasmgdb]
@@ -32,11 +33,13 @@ on crash to the user. When they are configured to emit a coredump, they collect
 the debugging information and write a coredump file.
 
 An example output:
+
 ```
-$ wasmrun module.wasm
+$ my-wasm-runtime module.wasm
 
 Exit 1: Uncaught RuntimeError: memory access out of bounds (core dumped).
 ```
+
 A coredump file has been generated.
 
 For experimenting, runtime support is not strictly necessary. A tool can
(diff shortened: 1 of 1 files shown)


fitzgen:
Gentle ping @dschuff

dschuff (commented):
Thanks for the ping. Overall I like this.

xtuc:
@fitzgen  naming doesn't make it obvious but the `core` section is used to describe the process, executable name, arguments, env variables etc. To start, it only contains the executable's name.

Coredump.md
```diff
@@ -86,11 +89,61 @@ As opposed to regular Wasm files, Wasm Coredumps are not [instantiated].
...
+
+This custom section must only appear once in the coredump file.
+
+This custom section establishes an index space of instances that can be
+referenced in `frame` productions within the `corestack` custom section (see
+below).
+
+```
+coreinstances ::= customsec(vec(coreinstance))
+coreinstance ::= 0x0 moduleidx:u32 memories:vec(u32) globals:vec(u32)
```
xtuc, line 129: should we restrict that a moduleidx must only appear once?

Coredump.md
```diff
@@ -99,7 +152,7 @@ debugging information.
 ```
 corestack   ::= customsec(thread-info vec(frame))
 thread-info ::= 0x0 thread-name:name
-frame       ::= 0x0 funcidx:u32 codeoffset:u32 locals:vec(value)
+frame       ::= 0x0 instanceidx:u32 funcidx:u32 codeoffset:u32 locals:vec(value)
```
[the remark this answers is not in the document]
xtuc: It's a breaking change for early tooling, which is fine :+1:

xtuc (approved):
Thanks @fitzgen

Coredump.md
fitzgen: This isn't a restriction we can enforce because if a module is instantiated multiple times, then we will need the same module index to appear multiple times in this section.

fitzgen:
> @fitzgen naming doesn't make it obvious but the `core` section is used to describe the process, executable name, arguments, env variables etc. To start, it only contains the executable's name.

Wasm itself doesn't really have these concepts, although WASI has them to a degree. I've filed https://github.com/WebAssembly/tool-conventions/issues/209 to further discuss this issue, since it doesn't need to block this PR.

PR #221, opened 2023-12-20
Signatures.md: update and document the signature identifier

jedisct1:
In order to avoid probing signatures that don't match a public key, signatures are prefixed with an algorithm identifier.

That identifier was present in examples, but not in the specification.

So, mention it in the specification.

In addition, current implementations don't consider that the identifier is part of the signature, so match these implementations. It doesn't make any difference beyond the encoded signature length.


jedisct1:
Ping?

PR #240, opened 2024-12-02
Document the C ABI for `_BitInt` and other types.

### LINKED ISSUE #239: Document Argument passing for _BitInt types
Forwarded from https://hachyderm.io/@kaylie@blahaj.zone/113574803870036780 .

The current BasicCABI.md does not document how _BitInt types are passed and it would be good to document it as such.

Note if someone does a GCC backend for WebAssembly, GCC folks in general has decided the ABI for _BitInt needs to be documented before it having it implemented too.

See https://github.com/ARM-software/abi-aa/issues/175, https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/300 and https://gitlab.com/x86-psABIs/x86-64-ABI/-/merge_requests/32 for examples on how other ABIs (have or will have document it).

sunfishcode:
Document the C ABI for `_BitInt`, `_Complex`, `__int128_t`, and `__uint128_t`. These follow the ABI currently used by clang, to the best of my knowledge.

Fixes #239.

### DIFF
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -38,6 +38,8 @@ General type | C Type | `sizeof` | Alignment (bytes) | Wasm Value Type
  Integer | `unsigned long` | 4 | 4 | i32
  Integer | `long long` / `signed long long` | 8 | 8 | i64
  Integer | `unsigned long long` | 8 | 8 | i64
+ Integer | `__int128_t` | 16 | 16 | (none)
+ Integer | `__uint128_t` | 16 | 16 | (none)
  Pointer | *`any-type *`* / *`any-type (*)()`* | 4 | 4 | i32
  Floating point | `float` | 4 | 4 | f32
  Floating point | `double` | 8 | 8 | f64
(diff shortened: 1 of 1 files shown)


sunfishcode:
It appears clang doesn't currently support `_BitInt` wider than 128 on any target other than x86 right now, so there isn't an existing ABI for it on Wasm that we can just document.

dschuff (approved):

sunfishcode:
The PR here includes documenting the alignment for `__int128_t`, which turns out to be interesting](https://github.com/rust-lang/rust/issues/133991#issuecomment-2527839550) because Clang defaults `__int128_t` to 16-byte alignment while LLVM defaults `i128` to 8-byte alignment. I've now posted https://github.com/llvm/llvm-project/pull/119204 to propose a fix for LLVM.

sunfishcode:
If anyone would like to have an ABI for `_BitInt(N)` where N > 128, they're welcome to file a new PR or issue to start the discussion. This PR is just about documenting existing practice.

PR #253, opened 2025-06-17
Fix typo in basic C ABI

zebreus:
The basic C ABI document mentions that `long long double` is being passed as two i64, when it should be just `long double`.

`long long double` does not exist.


sunfishcode:
Thanks!

PR #254, opened 2025-06-19
Add a Lime1 test file.

sunfishcode:
This PR adds a simple wast test file for the [Lime1 virtual CPU]. It is not a comprehensive conformance test, but it does at least mention all of Lime1's documented features, so it can be used as a quick way to determine if an engine supports the Lime1 feature set.

I've tested the test file with both Wasmtime and the [spec interpreter].

Some of the individual tests are derived from bytecodealliance/wasm-tools#2186.

[spec interpreter]: https://github.com/WebAssembly/spec/
[Lime1 virtual CPU]: https://github.com/WebAssembly/tool-conventions/blob/main/Lime.md#lime1


yamt (approved):
how about putting it under "test" directory? otherwise lgtm.

sunfishcode:
Merging, as this is just adding a test for the existing definition.

PR #258, opened 2025-10-18
Add text format specification for Linking.md

feedab1e:
This PR proposes a vendor-neutral syntax for describing relocation information in WAT.

Unlike the syntax described in https://github.com/WebAssembly/wabt/pull/2649, this proposal is intended to express everything that the binary format can, including whatever the current proposed implementation in WABT does not support.  Of note here is the fact that `@sym` annotations can appear multiple times per declaration, the inclusion of COMDATs, segment infos, and the inclusion of labels and addends.

### DIFF
--- a/Linking.md
+++ b/Linking.md
@@ -856,6 +856,15 @@ like following:
 
 ## Symbols
 
+For each relocatable WebAssembly entity type, there exists a corresponding
+symbol identifier namespaces for symbols of that type.
+
+Additionally, a symbol identifier namespace exists for data symbols.
+
+Symbol idenitfier namespaces differ from common index spaces in that they also
+allow purely textual names in addition to numeric + optional textual names
+allowed by index spaces.
+
 Symbols are represented as WebAssembly annotations of the form
 ```wat
 (@sym <name> <qualifier>*)
@@ -870,8 +879,16 @@ Data imports represented as WebAssembly annotations of the form
   symbol is considered *primary* symbol for that WebAssembly object, its name
   is taken from the related object
   - There may only be one primary symbol for each WebAssembly object.
-  - If a symbol is not associated with an object, it may not be the primary
-    symbol.
+  - If a symbol is not associated with a WebAssembly entity, it may not be the
+    primary symbol.
+
+After a name for the symbol is determined, it is placed into the symbol
+identifier namespace corresponding to that symbol type.
+
+> [!Note]
+> As a consequence of that, the only symbols that can be referred to by a
+> numeric index are _primary_ symbols, since they inherit their numeric index
+> form the relocatable WebAssebly object.
 
 - `qualifier` is one of the allowed qualifiers on a symbol declaration.
   Qualifiers may not repeat.
@@ -914,6 +931,11 @@ Shorthands may be used in place of full qualifiers:
 
 If all components of a symbol annotation are skipped, the annotation may be
 omitted.
+
+> [!Note]
+> Since all components of a symbol can be skipped, a _primary_ symbol always
+> exists for all WebAssembly entities, even if the annotation without a `name`
+> is not present in the symbol sequence
 
 ### WebAssembly object symbols
 
--- a/Linking.md
+++ b/Linking.md
@@ -769,6 +769,7 @@ Relocations are represented as WebAssembly annotations of the form
 | `datasec`   | data        | `R_WASM_SECTION_OFFSET`            | Offset into a data section        |
 | `customsec` | N/A         | `R_WASM_SECTION_OFFSET`            | Offset into a custom section      |
 | `data`      | data        | `R_WASM_MEMORY_ADDR_*`             | WebAssembly linear memory address |
+
 Symbol kinds marked with `*` are considered *primary*.
 
 - `modifier` describes the additional attributes that a relocation might have.
--- a/Linking.md
+++ b/Linking.md
@@ -764,9 +764,9 @@ Relocations are represented as WebAssembly annotations of the form
 | `global`    | global*     | `R_WASM_GLOBAL_INDEX_*`            | Final WebAssembly global index    |
 | `func`      | function*   | `R_WASM_FUNCTION_INDEX_*`          | Final WebAssembly function index  |
 | `functable` | function    | `R_WASM_TABLE_INDEX_*`             | Index into the dynamic function table, used for taking address of functions |
-| `text`      | function    | `R_WASM_FUNCTION_OFFSET`           | Offset into the function body     |
-| `funcsec`   | function    | `R_WASM_SECTION_OFFSET`            | Offset into a function section    |
-| `datasec`   | data        | `R_WASM_SECTION_OFFSET`            | Offset into a data section        |
+| `codeseg`   | function    | `R_WASM_FUNCTION_OFFSET`           | Offset into the function body from the start of the function |
+| `codesec`   | function    | `R_WASM_SECTION_OFFSET`            | Offset into the function section  |
+| `datasec`   | data        | `R_WASM_SECTION_OFFSET`            | Offset into the data section      |
 | `customsec` | N/A         | `R_WASM_SECTION_OFFSET`            | Offset into a custom section      |
 | `data`      | data        | `R_WASM_MEMORY_ADDR_*`             | WebAssembly linear memory address |
 
@@ -782,12 +782,12 @@ Symbol kinds marked with `*` are considered *primary*.
 
 - `addend` describes the additional components of a relocation.
 
-| `<addend>`   | interpretation       | condition                                                         |
-|--------------|----------------------|-------------------------------------------------------------------|
-| nothing      | Zero addend          | always                                                            |
-| `+<integer>` | Positive byte offset | `method` allows addend                                            |
-| `-<integer>` | Negative byte offset | `method` allows addend and `format` is signed                     |
-| `<labeluse>` | Byte offest to label | `method` allows addend and `method` is either `text` or `section` |
+| `<addend>`   | interpretation       | condition                                     |
+|--------------|----------------------|-----------------------------------------------|
+| nothing      | Zero addend          | always                                        |
+| `+<integer>` | Positive byte offset | `method` allows addend                        |
+| `-<integer>` | Negative byte offset | `method` allows addend and `format` is signed |
+| `<labeluse>` | Byte offest to label | `method` is either `codeseg` or `*sec`        |
 
 - `symbol` describes the symbol against which to perform relocation.
   - For `funcsec` relocation method, this is the function id, so that if the addend is zero, the relocation points to the first instruction of that function.
--- a/Linking.md
+++ b/Linking.md
@@ -899,7 +899,7 @@ identifier namespace corresponding to that symbol type.
 | `binding=<binding>`       | sets symbol flags according to `<binding>`    |
 | `visibility=<visibility>` | sets symbol flags according to `<visibility>` |
 | `retain`                  | sets `WASM_SYM_NO_STRIP` symbol flag          |
-| `thread_local`            | sets `WASM_SYM_TLS` symbol flag               |
+| `tls`                     | sets `WASM_SYM_TLS` symbol flag               |
 | `size=<int>`              | sets symbol's `size` appropriately            |
 | `offset=<int>`            | sets `WASM_SYM_ABSOLUTE` symbol flag, sets symbol's `offset` appropriately |
 | `name=<string>`           | sets `WASM_SYM_EXPLICIT_NAME` symbol flag, sets symbol's `name_len`, `name_data` appropriately |
--- a/Linking.md
+++ b/Linking.md
@@ -843,10 +843,11 @@ In data segments, relocation annotations can be interleaved into the data
 string sequence. When that happens, relocations are situated after the last
 byte of the value being relocated.
 
-For example, relocation of a 32-bit function pointer `$foo` into the data
-segment of size 4 would look like following:
+For example, relocation of a 32-bit function pointer `$foo` and a 32-bit
+reference to a data symbol `$bar` into the data segment of size 8 would look
+like following:
 ```wat
-(data (i32.const 0) "\00\00\00\00" (@reloc i32 functbl $foo))
+(data (i32.const 0) "\00\00\00\00" (@reloc i32 functbl $foo) "\00\00\00\00" (@reloc i32 data $bar))
 ```
 
 ## Symbols
--- a/Linking.md
+++ b/Linking.md
@@ -752,8 +752,8 @@ Relocations are represented as WebAssembly annotations of the form
 |`i64`     | `R_WASM_*_I64`                     | 8-byte [uint64]     |
 |`leb`     | `R_WASM_*_LEB`                     | 5-byte [varuint32]  |
 |`sleb`    | `R_WASM_*_SLEB`                    | 5-byte [varint32]   |
-|`leb64`   | `R_WASM_*_LEB`                     | 10-byte [varuint64] |
-|`sleb64`  | `R_WASM_*_SLEB`                    | 10-byte [varint64]  |
+|`leb64`   | `R_WASM_*_LEB64`                   | 10-byte [varuint64] |
+|`sleb64`  | `R_WASM_*_SLEB64`                  | 10-byte [varint64]  |
 
 - `method` describes the type of relocation, so what kind of symbol we are relocating against and how to interpret that symbol.
 
@@ -790,10 +790,10 @@ Symbol kinds marked with `*` are considered *primary*.
 | `<labeluse>` | Byte offest to label | `method` allows addend and `method` is either `text` or `section` |
 
 - `symbol` describes the symbol against which to perform relocation.
-  - For `funcsec` relocation metod, this is the function id, so that if the addend is zero, the relocation points to the first instruction of that function.
-  - For `datasec` relocation metod, this is the data segment id, so that if the addend is zero, the relocation points to the first byte of data in that segment.
-  - For `customsec` relocation metod, this is the name of the custom section, so that if the addend is zero, the relocation points to the first byte of data in that segment.
-  - For other relocation metods, this denotes the symbol in the scope of that symbol kind.
+  - For `funcsec` relocation method, this is the function id, so that if the addend is zero, the relocation points to the first instruction of that function.
+  - For `datasec` relocation method, this is the data segment id, so that if the addend is zero, the relocation points to the first byte of data in that segment.
+  - For `customsec` relocation method, this is the name of the custom section, so that if the addend is zero, the relocation points to the first byte of data in that segment.
+  - For other relocation methods, this denotes the symbol in the scope of that symbol kind.
 
 The relocation type is looked up from the combination of `format`, `method`, and `modifier`. If no relocation type exists, an error is raised.
 
@@ -807,11 +807,11 @@ If all components of a relocation annotation are skipped, the annotation may be 
 
 ### Instruction relocations
 
-For every usage of an `typeidx`, `funcidx`, `globalidx`, `tagidx`, a relocation annotation is added afterwards, with `format` predefined as `leb`, `method` predefined as the *primary* method for that type, and `symbol` defaulted as the *primary* symbol of that `idx`
+For every usage of `typeidx`, `funcidx`, `globalidx`, `tagidx`, a relocation annotation is added afterwards, with `format` predefined as `leb`, `method` predefined as the *primary* method for that type, and `symbol` defaulted as the *primary* symbol of that `idx`
 
-For the `i32.const` instruction, a relocation annotation is added after the integer literal operand, with `format` predefined as `sleb`, and `method` is allowed to be either `data` or `functable`.
-For the `i64.const` instruction, a relocation annotation is added after the integer literal operand, with `format` predefined as `sleb64`, and `method` is allowed to be either `data` or `functable`.
-For the `i{32,64}.{load,store}*` instructions, a relocation annotation is added after the offset operand, with `format` predefined as `leb` if the *memory* being referenced is 32-bit, and `leb64` otherwise, and `method` predefined as `data`.
+- For the `i32.const` instruction, a relocation annotation is added after the integer literal operand, with `format` predefined as `sleb`, and `method` is allowed to be either `data` or `functable`.
+- For the `i64.const` instruction, a relocation annotation is added after the integer literal operand, with `format` predefined as `sleb64`, and `method` is allowed to be either `data` or `functable`.
+- For the `i{32,64}.{load,store}*` instructions, a relocation annotation is added after the offset operand, with `format` predefined as `leb` if the *memory* being referenced is 32-bit, and `leb64` otherwise, and `method` predefined as `data`.
 
 ### Data relocations
 
@@ -838,29 +838,29 @@ Data imports represented as WebAssembly annotations of the form
 
 - `qualifier` is one of the allowed qualifiers on a symbol declaration. Qualifiers may not repeat.
 
-| `<qualifier>`    | effect                                                                           |
-|------------------|----------------------------------------------------------------------------------|
-| `weak`           | sets `WASM_SYM_BINDING_WEAK` symbol flag                                         |
-| `static`         | sets `WASM_SYM_BINDING_LOCAL` symbol flag                                        |
-| `hidden`         | sets `WASM_SYM_VISIBILITY_HIDDEN` symbol flag                                    |
-| `retain`         | sets `WASM_SYM_NO_STRIP` symbol flag                                             |
-| `thread_local`   | sets `WASM_SYM_TLS` symbol flag                                                  |
-| `size=<int>`     | sets symbol's `size` appropriately                                               |
-| `offset=<int>`   | sets `WASM_SYM_ABSOLUTE` symbol flag, sets symbol's `offset` appropriately       |
-| `name=<string>`  | sets `WASM_SYM_EXPLICIT_NAME` symbol flag, sets `name_len`, `name_data`          |
-| `priority=<int>` | adds symbol to `WASM_INIT_FUNCS` section with the given priority                 |
-| `comdat=<id>`    | adds symbol to a `comdat` with the given id                                      |
+| `<qualifier>`    | effect                                                                                         |
+|------------------|------------------------------------------------------------------------------------------------|
+| `weak`           | sets `WASM_SYM_BINDING_WEAK` symbol flag                                                       |
+| `static`         | sets `WASM_SYM_BINDING_LOCAL` symbol flag                                                      |
+| `hidden`         | sets `WASM_SYM_VISIBILITY_HIDDEN` symbol flag                                                  |
+| `retain`         | sets `WASM_SYM_NO_STRIP` symbol flag                                                           |
+| `thread_local`   | sets `WASM_SYM_TLS` symbol flag                                                                |
+| `size=<int>`     | sets symbol's `size` appropriately                                                             |
+| `offset=<int>`   | sets `WASM_SYM_ABSOLUTE` symbol flag, sets symbol's `offset` appropriately                     |
+| `name=<string>`  | sets `WASM_SYM_EXPLICIT_NAME` symbol flag, sets symbol's `name_len`, `name_data` appropriately |
+| `priority=<int>` | adds symbol to `WASM_INIT_FUNCS` section with the given priority                               |
+| `comdat=<id>`    | adds symbol to a `comdat` with the given id                                                    |
 
-`priority` qualifier may only be applied to function symbols.
-`size` and `offset` qualifiers may only be applied to data symbols.
-`size` and `name` qualifiers must be applied to data symbols.
-`name` qualifier must be applied to data imports.
+- The `priority` qualifier may only be applied to function symbols.
+- The `size` and `offset` qualifiers may only be applied to data symbols.
+- The `size` and `name` qualifiers must be applied to data symbols.
+- The `name` qualifier must be applied to data imports.
 
 If all components of a symbol annotation are skipped, the annotation may be omitted.
 
 ### WebAssembly object symbols
 
-For symbols related to WebAssembly objects, the sequence of symbol annotation occurs after the optional `id` of the declaration.
+For symbols related to WebAssembly objects, the symbol annotation sequence occurs after the optional `id` of the declaration.
 For example, the following code:
 ```wat
 (import "env" "foo" (func (@sym $a retain name="a") (@sym $b hidden name="b") (param) (result)))
@@ -870,7 +870,7 @@ declares 3 symbols: one primary symbol with the name of the index of the functio
 ### Data symbols
 
 Data symbol annotations can be interleaved into the data string sequence. When that happens, relocations are situated before the first byte of the value being defined.
-For example, declaration of a 32-bit global by with the name `$foo` and linkage name "foo" would look like following:
+For example, a declaration of a 32-bit global with the name `$foo` and linkage name "foo" would look like following:
 ```wat
 (data (i32.const 0) (@sym $foo name="foo" size=4) "\00\00\00\00")
 ```
@@ -891,22 +891,28 @@ COMDAT declarations occur in the same place as module fields.
 
 ## Labels
 
-For some relocation types, an offset into a section/function is necessary. For these cases, lablels exsist.
+For some relocation types, an offset into a section/function is necessary. For these cases, labels exsist.
 Labels are represented as WebAssembly annotations of the form
 ```wat
 (@sym.label <id>)
 ```
 
 ### Function labels
-Function labels occur in the same place as instructions. A label always denotes the first byte of the next instruction, or the byte after end of function's instruction stream if there isn't a next instruction
+Function labels occur in the same place as instructions.
+A label always denotes the first byte of the next instruction, or the byte after the end of the function's instruction stream, if there isn't a next instruction.
+
 Function label names are local to the function in which they occur.
 
 ### Data labels
-Data labels can be interleaved into the data string sequence. When that happens, relocations are situated after the last byte of the value being relocated.
+Data labels can be interleaved into the data string sequence.
+When that happens, relocations are situated after the last byte of the value being relocated.
+
 Data label names are local to the data segment in which they occur.
 
 ### Custom labels
-Custom labels can be interleaved into the data string sequence. When that happens, relocations are situated after the last byte of the value being relocated.
+Custom labels can be interleaved into the data string sequence.
+When that happens, relocations are situated after the last byte of the value being relocated.
+
 Custom label names are local to the custom section in which they occur.
 
 ## Data segment flags
@@ -917,13 +923,13 @@ Data segment flags are represented as WebAssembly annotations of the form
 
 - `qualifier` is one of the allowed qualifiers on a data segment declaration. Qualifiers may not repeat.
 
-| `<qualifier>`   | effect                                        |
-|-----------------|-----------------------------------------------|
-| `align=<int>`   | sets segment's `alignment` appropriately      |
-| `name=<string>` | sets `name_len`, `name_data`                  |
-| `strings`       | sets `WASM_SEGMENT_FLAG_STRINGS` segment flag |
-| `thread_local`  | sets `WASM_SEGMENT_FLAG_TLS` segment flag     |
-| `retain`        | sets `WASM_SEG_FLAG_RETAIN` segment flag      |
+| `<qualifier>`   | effect                                               |
+|-----------------|------------------------------------------------------|
+| `align=<int>`   | sets segment's `alignment` appropriately             |
+| `name=<string>` | sets segment's `name_len`, `name_data` appropriately |
+| `strings`       | sets `WASM_SEGMENT_FLAG_STRINGS` segment flag        |
+| `thread_local`  | sets `WASM_SEGMENT_FLAG_TLS` segment flag            |
+| `retain`        | sets `WASM_SEG_FLAG_RETAIN` segment flag             |
 
 If `align` is not specified, it is given a default value of 1.
 If `name` is not specified, it is given an empty default value.
--- a/Linking.md
+++ b/Linking.md
@@ -290,11 +290,12 @@ where a `syminfo` is encoded as:
 |              |                |   `5 / SYMTAB_TABLE`                        |
 | flags        | `varuint32`    | a bitfield containing flags for this symbol |
 
-For functions, globals, events and tables, we reference an existing Wasm object, which
-is either an import or a defined function/global/event/table (recall that the operand of a
-Wasm `call` instruction uses an index space consisting of the function imports
-followed by the defined functions, and similarly `get_global` for global imports
-and definitions and `throw` for event imports and definitions).
+For functions, globals, events and tables, we reference an existing WebAssembly
+entity, which is either an import or a defined function/global/event/table
+(recall that the operand of a Wasm `call` instruction uses an index space
+consisting of the function imports followed by the defined functions, and
+similarly `get_global` for global imports and definitions and `throw` for event
+imports and definitions).
 
 If a symbols refers to an import, and the
 `WASM_SYM_EXPLICIT_NAME` flag is not set, then the name is taken from the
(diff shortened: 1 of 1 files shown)


Linking.md
```diff
@@ -734,3 +734,206 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+Symbols are represented as WebAssembly annotations of the form
+```wat
+(@sym <name> <qualifier>*)
+```
+Data imports represented as WebAssembly annotations of the form
+```wat
+(@sym.import.data <name> <qualifier>*)
+```
+
+- `name` is the symbol name written as WebAssembly `id`, it is the name by which relocation annotations reference the symbol. If it is not present, the symbol is considered *primary* symbol for that WebAssembly object, its name is taken from the related object
```
sbc100: Can you warp long lines like this at 80 columns

Linking.md
```diff
@@ -734,3 +734,206 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+  - There may only be one primary symbol for each WebAssembly object.
+  - If a symbol is not associated with an object, it may not be the primary symbol.
+
+- `qualifier` is one of the allowed qualifiers on a symbol declaration. Qualifiers may not repeat.
+
+| `<qualifier>`    | effect                                                                                         |
+|------------------|------------------------------------------------------------------------------------------------|
+| `weak`           | sets `WASM_SYM_BINDING_WEAK` symbol flag                                                       |
+| `static`         | sets `WASM_SYM_BINDING_LOCAL` symbol flag                                                      |
```
sbc100: `static` is more of a C language term isn't it?   At the object file level i think `local` and `global` is the terminology isn't it?   Also, would we want to be future proof against other binding types?  e.g should we use `binding=local` here?  That sound seem to correspond more directly the the linking section too.

Linking.md
```diff
@@ -734,3 +734,206 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+| `hidden`         | sets `WASM_SYM_VISIBILITY_HIDDEN` symbol flag                                                  |
+| `retain`         | sets `WASM_SYM_NO_STRIP` symbol flag                                                           |
```
sbc100: If we think retain is a better terminology then we should probably propose renaming `WASM_SYM_NO_STRIP` rather than diverging.

Linking.md
```diff
@@ -734,3 +734,206 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+(import "env" "foo" (func (@sym $a retain name="a") (@sym $b hidden name="b") (param) (result)))
+```
+declares 3 symbols: one primary symbol with the name of the index of the function, one symbol with the name `$a`, and one symbol with the name `$b`.
+
+### Data symbols
+
+Data symbol annotations can be interleaved into the data string sequence. When that happens, relocations are situated before the first byte of the value being defined.
+For example, a declaration of a 32-bit global with the name `$foo` and linkage name "foo" would look like following:
+```wat
+(data (i32.const 0) (@sym $foo name="foo" size=4) "\00\00\00\00")
```
sbc100: Can we perhaps show the interleaving in the example, so maybe include two relocations here?

Linking.md
```diff
@@ -734,3 +734,206 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
 instruction) do not yet exist.
 - There is currently no support for table element segments, either active or
 passive.
+
+# Text format
+
```
sbc100: Perhaps add some high level info here.  something like "The linking metadata section can be represented in the text format at a set of WebAssembly annotations.  Tooling that supports these annotation can be used to round trip a WebAssembly object file through the wat format."

sbc100 (commented):
Seems like a good idea to me.  Good to see the annotation proposal being used in innovative ways like this!

Linking.md
feedab1e: Yeah, that's sound. I just thought that those names are from LLVM's source, so that rename should be done in sync with LLVM

Linking.md
feedab1e: Added

Linking.md
feedab1e: Fixed

sbc100:
For comparison I wonder if you could post a simple `hello_world.o` object file in both `.s` assembly format and the `.wat` format.   We should probably almost mention that LLVM uses its own `.s` format somewhere here too, but that can be separate PR if you like.

Linking.md
feedab1e: Added

feedab1e:
On second thought, I can just use `clang -S` for `.s` and `clang -c` `wasm2wat` for `.wat`, so I can do that, yeah

feedab1e:
@sbc100 My "hello world" examples turned out to be more than a full screen of code when using `.s`. I can send the files here, but I don't think it would be good to put that output into the document.

feedab1e:
Oh, sure, then, for the source file [test.cpp](https://github.com/user-attachments/files/22987863/test.cpp), when compiled with `clang -S -O3`, yields [test.s](https://github.com/user-attachments/files/22987866/test.s.txt), and when compiled  `clang -c -O3`, and than run through `wasm2wat -r --enable-annotations` (provided wabt from my fork), yields [test.wat](https://github.com/user-attachments/files/22987872/test.wat.txt).

(Unfortunately, github forbids me from posting `.s` and `.wat`, so `.txt` it is)

feedab1e:
One thing I wanted to mention here is that there is a limitation in the text format, that code section relocations that do not make sense for instruction operands may not be expressed in the text format. So for example `i32.load` instruction with a `R_WASM_TYPE_INDEX_LEB` could occur in a perfectly valid object file, from binary format's PoV, yet would be inexpressible in the text format. Do we want to disallow such relocations in the binary format too?
(AFAIK this would also be a problem for `.s`)

feedab1e:
Would you be fine with me adding into the doc that such relocations are invalid for the purposes of validation, then?
And same for `R_WASM_SECTION_OFFSET_I32` addends.
Also, do we want to specify that section entries in relocation tables may only point to the data section, the code section, or custom sections?

Linking.md
```diff
@@ -181,6 +177,51 @@ relocations applied to the CODE section, a relocation cannot straddle two
...
+files.
+
+All LEB128-encoded values that are to be relocated must be maximally padded so
+that they can be rewritten without affecting the position of any other bytes.
+For instance, the function index 3 must be encoded as `0x83 0x80 0x80 0x80 0x00`.
+
+When relocations occur in the CODE section, only the following relocations may
+occur:
+
+| relocation type                 | condition the value at relocation offset |
```
sbc100: How about "target" instead of "offset"?

Linking.md
```diff
@@ -322,6 +363,8 @@ For section symbols:
 | ------------ | -------------- | ------------------------------------------- |
 | section      | `varuint32`    | the index of the target section             |
 
+Section symbols may only reference the CODE section, the DATA section, or custom sections.
```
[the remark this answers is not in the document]
feedab1e: I don't believe it can't work like that, though, since `R_WASM_SECTION_OFFSET_I32` reference a section symbol, and for that to work as DWARF code addresses, the symbol that relocation references would have to reference the CODE section, while the relocation itself would have to target a place in the debug section.

Linking.md
```diff
@@ -181,6 +177,51 @@ relocations applied to the CODE section, a relocation cannot straddle two
...
+the addend:
+
+- If `index` references the CODE section, the addend must represent the first
+  byte of an instruction, or the byte after the last instruction.
+- If `index` references the DATA section, the addend must represent a valid
+  offset into a data segment's data area.
+- If `index` references the custom section, the addend must represent a valid
+  offset into that custom section's data area.
+
+All other relocations are considered invalid for the purposes of validation
```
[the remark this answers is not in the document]
feedab1e: Did that, https://github.com/WebAssembly/tool-conventions/pull/259, please take a look.

Linking.md
```diff
@@ -181,6 +177,51 @@ relocations applied to the CODE section, a relocation cannot straddle two
 functions, and for the DATA section relocations must lie within a data element's
 body.
 
+### Additional validation rules
+
+When perfoming validation on object files, care must be taken to ensure that
+meaningless relocations are not present in the binary.
```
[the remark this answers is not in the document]
sbc100: I don't think custom sections are allowed to cause the normal "Wasm validation" to fail, but we can define our own higher level notions of a valid object file on top I believe.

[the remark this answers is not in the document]
sbc100: SGTM.  

This also fits with native linker model which seems to have a kind of garbage-in, garbage-out model for input object and relocations.

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+| `tls`        | `R_WASM_*_TLS*`                       | Address relative to `env.__tls_base`, used for thread-local storage |
+
+- `addend` describes the additional components of a relocation.
+
+| `<addend>`   | interpretation       | condition                                     |
+|--------------|----------------------|-----------------------------------------------|
+| nothing      | Zero addend          | always                                        |
+| `+<integer>` | Positive byte offset | `method` allows addend                        |
+| `-<integer>` | Negative byte offset | `method` allows addend and `format` is signed |
+| `<labeluse>` | Byte offest to label | `method` is either `codeseg` or `*sec`        |
```
alexcrichton: What does `<labeluse>` correspond to in https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md? I would expect this to be an integer-or-not, and I'd be worried about a bare integer 1 being confused with "+1" for example given the current syntax.

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+- For the `i32.const` instruction, a relocation annotation is added after the
+  integer literal operand, with `format` predefined as `sleb`, and `method` is
+  allowed to be either `data` or `functable`.
+- For the `i64.const` instruction, a relocation annotation is added after the
+  integer literal operand, with `format` predefined as `sleb64`, and `method`
+  is allowed to be either `data` or `functable`.
+- For the `i{32,64}.{load,store}*` instructions, a relocation annotation is
+  added after the offset operand, with `format` predefined as `leb` if the
+  *memory* being referenced is 32-bit, and `leb64` otherwise, and `method`
+  predefined as `data`.
```
alexcrichton: Would it be easier to say that relocations always occur after an instruction and avoid talking about each variant individually here?

Another question: how come relocations are rendered after instructions vs before? Same for data segments where the relocation comes after what's being relocated.

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+
+- `qualifier` is one of the allowed qualifiers on a symbol declaration.
+  Qualifiers may not repeat.
+
+| `<qualifier>`             | effect                                        |
+|---------------------------|-----------------------------------------------|
+| `binding=<binding>`       | sets symbol flags according to `<binding>`    |
+| `visibility=<visibility>` | sets symbol flags according to `<visibility>` |
+| `retain`                  | sets `WASM_SYM_NO_STRIP` symbol flag          |
+| `thread_local`            | sets `WASM_SYM_TLS` symbol flag               |
```
alexcrichton: I thought I saw a comment go by about this, but wanted to add my thoughts on this as well: I think that the C contants should match what's in the text format. Whether the C name changes or the text name changes I don't mind, but similar to the "layers of abstraction" comment I had above I don't think it's worthwhile to invent new names just for the text format.

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+| `size=<int>`              | sets symbol's `size` appropriately            |
+| `offset=<int>`            | sets `WASM_SYM_ABSOLUTE` symbol flag, sets symbol's `offset` appropriately |
+| `name=<string>`           | sets `WASM_SYM_EXPLICIT_NAME` symbol flag, sets symbol's `name_len`, `name_data` appropriately |
+| `priority=<int>`          | adds symbol to `WASM_INIT_FUNCS` section with the given priority |
+| `comdat=<id>`             | adds symbol to a `comdat` with the given id   |
```
alexcrichton: Assuming that `<id>` can be something like `$foo` this is typically done in the text format as something like `(comdat $foo)` instead of `comdat=$foo` since I think the latter tokenizes as a single token rather than already being pre-split

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+| `default`      |                              |
+| `hidden`       | `WASM_SYM_VISIBILITY_HIDDEN` |
+
+Shorthands may be used in place of full qualifiers:
+
+| shorthand | resulting qualifier |
+|-----------|---------------------|
+| `hidden`  | `visibility=hidden` |
+| `local`   | `binding=local`     |
+| `weak`    | `binding=weak`      |
```
alexcrichton: Given these shorthands is there any need for `binding=<binding>` or `visibility=<visibility>` when these could be used instead?

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+| `<qualifier>`             | effect                                        |
+|---------------------------|-----------------------------------------------|
+| `binding=<binding>`       | sets symbol flags according to `<binding>`    |
+| `visibility=<visibility>` | sets symbol flags according to `<visibility>` |
+| `retain`                  | sets `WASM_SYM_NO_STRIP` symbol flag          |
+| `thread_local`            | sets `WASM_SYM_TLS` symbol flag               |
+| `size=<int>`              | sets symbol's `size` appropriately            |
+| `offset=<int>`            | sets `WASM_SYM_ABSOLUTE` symbol flag, sets symbol's `offset` appropriately |
+| `name=<string>`           | sets `WASM_SYM_EXPLICIT_NAME` symbol flag, sets symbol's `name_len`, `name_data` appropriately |
+| `priority=<int>`          | adds symbol to `WASM_INIT_FUNCS` section with the given priority |
```
alexcrichton: Perhaps `init-func-priority` or something to help indicate that this priority here is related to initialization?

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+
+For symbols related to WebAssembly objects, the symbol annotation sequence
+occurs after the optional `id` of the declaration.
+
+For example, the following code:
+```wat
+(import "env" "foo" (func (@sym $a retain name="a") (@sym $b hidden name="b") (param) (result)))
+```
+declares 3 symbols: one primary symbol with the name of the index of the
+function, one symbol with the name `$a`, and one symbol with the name `$b`.
```
alexcrichton: Can you clarify what the "primary symbol" is here? As it relates to linking.md I would expect this to introduce 2 symbols, only `$a` and `$b`, and I'm not sure what "primary symbol" is in this case.

alexcrichton (commented):
Thanks for this! I'd second the idea of including examples directly in this document if you're up for it. Would it be possible to show the C/C++ source you're working with plus the raw text output? Maybe not the entire file in all cases but it'd be convenient to see example usage of all of these dierctives within the context of actual `*.wat`.

sbc100:
> Thanks for this! I'd second the idea of including examples directly in this document if you're up for it. Would it be possible to show the C/C++ source you're working with plus the raw text output? Maybe not the entire file in all cases but it'd be convenient to see example usage of all of these dierctives within the context of actual `*.wat`.

I think a simple hello world in C should be enough since that will include a function relocation for `printf` and a data relocation for the `hello world` string.   No need to do anything more fancy in the basic example.

Linking.md
feedab1e: Hmm, actually, would it be better to write all qualifiers like that, or even support both styles?

Linking.md
feedab1e: > Would it be easier to say that relocations always occur after an instruction and avoid talking about each variant individually here?

While I don't think that there are currently cases where that happens, in the future there there will be a need for multiple relocations per instruction (for example, in multimemory), where one will have to write something like `i32.load 0 (@reloc $mymem) offest=0 (@reloc $my_global_var)`, so the first relocation is actually in the middle of an instruction, not after it.

Another reason for this is that I still need to predefine the relocation methods here so that the relocations are more brief (so one write `i32.load (@reloc $foo`) instead of `i32.load (@reloc leb data $foo)`).

> Another question: how come relocations are rendered after instructions vs before? Same for data segments where the relocation comes after what's being relocated.

Preference: Basically, `i32.const 0 (@reloc functable $foo)` looks better to my eyes than `i32.const (@reloc functable $foo) 0`. Data relocations are placed the same way for consistency, so that the user doesn't have to trip over the areas of effect being different.

Linking.md
feedab1e: Well, I used _entity_ before I saw _object_ used in the same context in the spec, so if we don't like _object_ as a term, I'd propose using that. The problem with _element_ is that there is a WebAssembly meaning behind this term, so we can't use it either.

sbc100: entity sounds better than object to me yes.   Naming is hard.

feedab1e:
As for disassembling a "hello world" compiled with `-g`, the problem here is that I can't disassemble those without information loss using WABT, just like I can't properly validate that such a binary is a valid object file. This is a limitation in the current design of WABT's binary reader, where I don't have up-to-date information on the file offsets of each operand and/or instruction, as well as binary reader being single-pass in combination with relocation section occurring later than the section being relocated.

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+
+The text format for linking metadata is intended for WAT consumers that wish to
+emit relocatable object files, and WAT producers wish to emit human-readable
+relocation metadata for later creation of a relocatable object file.
+
+## Relocations
+
+Relocations are represented as WebAssembly annotations of the form
+```wat
+(@reloc <format> <method> <modifier> <symbol-reference> <addend>)
```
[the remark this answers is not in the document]
feedab1e: There is also an issue of `R_WASM_TABLE_INDEX_*` vs `R_WASM_TABLE_NUMBER_*`, where clearly the choice of `R_WASM_TABLE_INDEX_*` for an index into a table was wrong and clashed with the relocation type for an index of a table, and I don't want to cement that mistake even further into the format.

Rectifying that error at the source would require me to patch LLVM in sync with this change, so like the other issue with `WASM_SYM_NO_STRIP`/`WASM_SEG_FLAG_RETAIN`/`retain` that @sbc100 raised in https://github.com/WebAssembly/tool-conventions/pull/258#discussion_r2442498877, I think this should be split into a different PR and handled later.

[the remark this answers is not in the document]
sbc100: Do you have a better idea for what to call `R_WASM_TABLE_INDEX_*`?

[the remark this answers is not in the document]
sbc100: (By the way the history here is that prior to mulit-table there was only one table, so `R_WASM_TABLE_INDEX_` was unambiguous when it was created.   I still think its reasonable to think "what is the table index for a given function", meaning the index in the table at which that function lives, but I'm open a better ideas).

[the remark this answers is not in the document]
sbc100: Maybe R_WASM_TABLE_OFFSET_?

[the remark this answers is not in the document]
feedab1e: (coincidentally this would align well with `functable` relocation method name)

Linking.md
feedab1e: Now using _entity_ everywhere where that refers to a WebAssembly entity

feedab1e:
> > > When relocation entries are implicit like this how does the binary writer know to produce them? Would there be special flag to the wat2wasm program that says something like `--generate-object-file` / `--add-implicit-relocs` ? Or would it be somehow automatic? i.e. what would stop the wat2wasm program generating relocs in normal non-object-file cases?
> > 
> > 
> > Yes, that flag already exists in `wat2wasm` trunk, it's `-r`, or `--relocatable`
> 
> Do you think this a good design (its something I just threw together back in the day)?

Well, it does make the code prettier, and it does allow to link against all existing WAT modules with no source changes, I see that as a benefit.

> I wonder if "explicit relocs everywhere" might not be better? Otherwise won't all the tools that do text to binary conversion will need some kind of extra flag to see if implicit relocs are enabled or not?
> 
> Or can we magically enable implicit relocs whenever we see any kind linking annotation in the wat file? Could there exist a wat file with zero explict annotations? How would the tooling know to create a linking section or not in that case?

Well, if we really do want that, I suppose we could dispatch based on whether the features section is present. But that wouldn't be reliable in any case, since tooling that isn't relocation-aware would skip unknown annotations and silently yield a simple module instead of an object file.

Linking.md
alexcrichton: If there's not too strong of a preference any which way I'd mid-ways-strongly-prefer to drop the `binding=...` and `visibility=...` annotations in favor of just the bare shorthand flags. That way `binding=global` wouldn't be a thing to confuse with wasm globals and additionally the linker names are already there as `local` for example instead of `static`.

Linking.md
alexcrichton: Ok so to make sure I understand this right: when this says here that 3 symbols are created, those are text-level-parsing symbols? This is in contrast to the linking-section-symbols of which only 2 are created here, right? (speaking of overloaded terminology...). 

If this is correct I might recommend reserving the term "symbol" for things related to the `linking` section and using "identifier" to mean "this is a `$foo` thing in the text format to have the text format be more human readable"

Linking.md
```diff
@@ -734,3 +734,279 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+### Data symbols
+
+Data symbol annotations can be interleaved into the data string sequence.
+When that happens, relocations are situated before the first byte of the value
+being defined.
+
+For example, a declaration of a 32-bit global with the name `$foo` and linkage
+name "foo" would look like following:
+```wat
+(data (i32.const 0) (@sym $foo name="foo" size=4) "\00\00\00\00")
```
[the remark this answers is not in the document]
alexcrichton: Sugar perhaps? `(@sym $foo size=4)` implicitly has `name="foo"` and `(@sym $foo2 name="foo" size=4)` would be a duplicately-named-symbol but text-format-unique-symbol

Linking.md
feedab1e: Ok, fixed

Linking.md
feedab1e: Honestly, I also do have a bit of a preference towards the shorthands, and considering that with the paren-syntax it would look like `(@sym (binding global))`, I don't actually feel that this will bring more clarity over just `(@sym global)`. I don't see a future where any of the options in `binding` and `visibility` would clash, either, so that's also not an issue.

Linking.md
feedab1e: Using `init_prio` now, hope that's clearer without being too long.

sbc100:
> > > I wonder if "explicit relocs everywhere" might not be better? Otherwise won't all the tools that do text to binary conversion will need some kind of extra flag to see if implicit relocs are enabled or not?
> > 
> > 
> > I may not fully be following the context here, but my hope is that `wasm-tools print` wouldn't need any sort of special flags to print `@reloc` annotations from a `*.o` file. Similarly I wouldn't want to need to pass extra flags to `wasm-tools parse` to generate `reloc.*` and `linking` sections when generating a `*.o` file.
> 
> So, @sbc100 is basically asking if it would be possible at all to assemble a file form WAT without emitting relocation metadata. For disassembly it'd be easy, since it's always obvious if a file is an object file based on the presence of the relevant custom sections. But for assembly with current syntax, any valid text module can also be an object file now.

I suppose we could make a rule such as "If at least one explicit linking annotation exists in the wat file then the whole file is assumed to be relocatable, and implicit relocations will be injected/generated in all possible locations"

Should we go with "at least on explicit annotation" or should we maybe have some kind of top level annotation that expresses "this is a relocatable object file"?

sbc100:
> Looking at prior art, if I am to run `wat2wasm --enable-code-metadata --enable-annotations` on a file with no code metadata annotations, do the sections still appear?

I think they would appear in the binary if and only if they appear in the source (wat).

The idea of creating implicit annotations (i.e. annotation that don't exist in the wat file at all) is the tricky thing here.  I'm not sure its good idea to that route.

alexcrichton:
Personally I would advocate for `linking` and `reloc.*` sections are only generated if the input has `@sym` or `@reloc` by default. I'd prefer to add more flags to `wasm-tools parse`-the-CLI (the text-to-binary conversion) as that would also proliferate to all API users of the programmatic functionality as well. To me it also feels less magic to have annotations-per-item as opposed to generating them automatically

Linking.md
sbc100: I tend to agree with @alexcrichton here (assuming I understand the above correctly).

I think we should perhaps not allow the omitting the annotations.   Maybe we just require them in all cases, at least initially for the first draft.   That way, the contents of the linker section and symbol table much more obviously corresponds 1-to-1 with the annotations in the wat.

if we  later decide it really is too verbose we could always add something later, some way to enable eliding of annotations.   I can imagine several ways to opt into that more magical behaviour, but perhaps we should start with a the simpler, explicit version?

Linking.md
feedab1e: Well, those aren't described in trunk, but are described here: https://github.com/feedab1e/tool-conventions/blob/main/Linking.md#labels.
They are basically used in conjunction with `R_WASM_SECTION_OFFSET_*` to represent [DWARF code address](https://yurydelendik.github.io/webassembly-dwarf/#pc) relocations.

You stick this label into the instruction stream, and whenever that label is used as an addend, it will point to that place in the instruction stream.
```wat
(module
  (func $foo (param) (result)
    nop
    (@sym.label $between_nops)
    nop)
  (@custom "debug_whatever"
    (;suppose this would represent a DWARF code address;)
    "\00\00\00\00\00" (@reloc leb codesec $foo $between_nops)))
```

Linking.md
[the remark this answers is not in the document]
alexcrichton: Oh I naively thought that all symbols required a name, but if that's not true then yeah `name=...` would be required.

alexcrichton:
My assumption is that the presence of `@reloc` and `@sym` causes stuff to be generated internally, but otherwise nothing happens. For example `call $foo` wouldn't generate any relocations by default (I realize this is against the `-r` flag y'all seem to be describing with wabt). So I was naively thinking that nothing would need stripping, only things would need to be preserved if they were present.

Linking.md
feedab1e: > You'll have to forgive me if these questions I'm raising seem naive. I am not intimately familiar with the `linking` and `reloc.*` sections, nor exactly how LLVM uses all the bits and bobs everywhere. I'm familar with everything at a high-level, however, and much of my thinking is projected from an assumed understanding of how things work.
> 
> One thing I've never understood is how symbols in the linker-sense related to the name section in the text-format sense. For example is the symbol of `(import "a" "b" (func $bar))` `"a"`, `"b"`, `"bar"`, or just the index 0? Or does `"a"` have to be `"env"` and nothing else?

My understanding is that the linking spec is entirely unaware of the `name` section, and the only strings that participate in deciding on the linkage name for a symbol are the import's `field` and the explicit name, if present.

> Another observation/thought, related to the discussion below, is that I would be surprised at the ability to omit `(@sym)`. For example I would expect that the text-to-binary conversion for the module you pasted would not result in a `linking` or `reloc.*` section. I would assume that `(@sym)` and `(@reloc ...)` are required to get something into the custom sections, and otherwise nothing is emitted.

> I think we should perhaps not allow the omitting the annotations. Maybe we just require them in all cases, at least initially for the first draft. That way, the contents of the linker section and symbol table much more obviously corresponds 1-to-1 with the annotations in the wat.

Sure, I can disable omission for just `(@sym)`.

> if we later decide it really is too verbose we could always add something later, some way to enable eliding of annotations. I can imagine several ways to opt into that more magical behaviour, but perhaps we should start with a the simpler, explicit version?

The main thing I'm concerned about is that that will still break WABT users that relied on this magic, but on the other hand I can't imagine that that many people relied on that feature in the first place, so it's probably fine.

Linking.md
feedab1e: > Aha ok I see! Two thoughts then. Maybe `@sym.bind` instead of `@sym.label` to avoid overloading terminology with the wasm labels?

Sure, I can do that.

> Also, in terms of ambiguity in parsing, what I was worried about was something like `(@sym.label)` (without a `$foo`-name) introducing symbol "0" and then the relocation wants to refer to it and 0 would ambiguously be an offset of 0 bytes or the 0th symbol. As I write this out though this seems silly. The specification could instead be that all symbols are `$`-bound, right?

And that's how the spec works currently, I believe.

> (and if so could "1" be allowed as an offset to avoid needing "+1"?)

Perhaps, but I think that a `+` would more clearly signal that this is an offset in relation to a symbol (so, `symbol +offset` looks more like an expression), rather than just any old number (which can also be mistaken for an index). So in context that would be `(@reloc i32 data $datasym +2)` vs `(@reloc i32 data $datasym 2)`.

alexcrichton: For "bind" terminology to me what's happening is that `(@sym.bind $foo)` is binding the symbol `$foo` to the current location of wherever the annotations is placed, so a dual definition operation of defining the symbol `$foo` and simultaneously binding it to the specified location. 

Although I suppose a possible alternative is just `(@sym $foo)` without using label terminology either? I mostly just want to avoid confusion with wasm labels.

For `+1` vs `1`, what about `(@reloc i32 data $datasym (offset 2))`, (@reloc i32 data $datasym (offset -2))`, and maybe (@reloc i32 data $datasym (offset-sym $foo))`? (this is really bikesheddy so the spec IMO can also just sort of do whatever here)

feedab1e:
> Could handling func-with-`@sym` plus func-without-`@sym` be part of the validation of the linking section? Where validation requries that if any entity has `@sym` then they all have `@sym`? And then dealing with the text format of invalid modules is expected to return an error of some kind?

Sure, but how would you implement those diagnostics? The parser state is gone by the time you do object file validation, as well as source locations for the declarations.

feedab1e:
> > I guess my actual question is how would you diagnose the following:
> > ```webassembly
> > (module
> >   (func (param) (result))
> >   (func (@sym) (param) (result)))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is either
> > 
> > 1. "missing `(@sym)`" on the first declaration (all WebAssembly entities are required to have at least one symbol)
> > 2. "`(@sym)` is not allowed" on the second declaration (upon seeing the first declaration, we decided that we aren't actually processing an object file)
> > 
> > Option 1 would require either restarting the parse, or remembering all of the declaration locations, and then issuing an error for every one that had no annotations.
> 
> How about we require `@relocatable` on the module itself?

That would work, too. Actually, then, I think it would be better to have an annotation describing all WAT features that are enabled for the module.
Something like:
```wat
(module (@wat-features linking code-metadata dwarf))
```

For object file linking specifically, another option is to say something like
```wat
(module (@target-features +overlong-leb -multimemory))
```
to trigger object file generation, since those make sense only for linking, AFAIK

Linking.md
feedab1e: > For "bind" terminology to me what's happening is that `(@sym.bind $foo)` is binding the symbol `$foo` to the current location of wherever the annotations is placed, so a dual definition operation of defining the symbol `$foo` and simultaneously binding it to the specified location.

Yeah, I don't really feel good using "bind" there, since any use of a dollar name in any definition "binds" that name to that definition, so it's not really descriptive of *what* is bound in this case.

> Although I suppose a possible alternative is just `(@sym $foo)` without using label terminology either? I mostly just want to avoid confusion with wasm labels.

I think, then, that `(@sym.loc)` would fit best, since that's what they're used for locations in the binary and are not conflicting with wasm labels.

> For `+1` vs `1`, what about `(@reloc i32 data $datasym (offset 2))`, (@Reloc i32 data $datasym (offset -2))`, and maybe (@reloc i32 data $datasym (offset-sym $foo))`? (this is really bikesheddy so the spec IMO can also just sort of do whatever here)

Well, we do have `offset=<num>` as part of a `memarg`, can we use that here too?

sbc100: I think both forms are valid (I assume they are both generated by llvm today).  Its up to the compiler to perform these optimizations (constant folding?) but we shouldn't mandate it one or the other here.

sbc100: IIRC there are some cases where this folds cannot always be performed.   I vaguely remember something about the `offset` being limited to positive values causing issues, but I can't remember the exact issue.  

There was some recent work in llvm to make more use of the offset: https://github.com/llvm/llvm-project/pull/145829

feedab1e: > I think both forms are valid (I assume they are both generated by llvm today). Its up to the compiler to perform these optimizations (constant folding?) but we shouldn't mandate it one or the other here.

No, this is only concerning WAT notation for the instructions, no codegen changes. 

Relocation on a `memarg` replaces the integer that would otherwise be specified as `offset=<num>`, so this number is basically unused for purposes of linking.

A potential issue would be in that this number is still specified in the main binary, so currently `i32.load offset=3 (@reloc $datasym +5)` is representable, but may not be if that change is adopted. A solution to that would be to default instead of predefining the addend component for the `memarg` case

Linking.md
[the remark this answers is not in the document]
feedab1e: No, actually, that can absolutely work if `WASM_*_OFFSET_*` relocations actually resolve to offsets form the file start, like DWARF actually expects. I do think the current spec is not very clear on this and someone from LLVM should take a look at what actually happens there and adjust https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#processing-relocations accordingly.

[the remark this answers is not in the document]
feedab1e: Updated the text format to reflect that section offset relocations may only reference custom sections for now.

sbc100 (commented):
Don't we still need to resolve the issue of how wat reader knows if it should produce a relocatable output at all?

My current preference is for a global annotation on the top level module, but I don't see that reflected in the current PR.

I'm also still not sure about the implicit relocation for things like `call $foo` .. does that currently generate an implicit relocation?    I still think it might be good to remove all implicit relocation for v1 of the spec, although the addition of the the level annotation on the module object makes me feel a little better about it.

I'm curious if other folks have opinions about implicit relocations.  @dschuff maybe?  @sunfishcode @tlively ?    I guess it kind of matches assembly formats where `call foo` is an implicit relocation.

Linking.md
```diff
@@ -734,3 +735,288 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+
+Symbol kinds marked with `*` are considered *primary*.
+
+- `modifier` describes the additional attributes that a relocation might have.
+
+| `<modifier>` | corresponding relocation constants    | interpretation    |
+|--------------|---------------------------------------|-------------------|
+| nothing      | nothing                               | Normal relocation |
+| `pic`        | `R_WASM_*_LOCREL_*`, `R_WASM_*_REL_*` | Address relative to `env.__memory_base` or `env.__table_base`, used for dynamic linking |
+| `tls`        | `R_WASM_*_TLS*`                       | Address relative to `env.__tls_base`, used for thread-local storage |
```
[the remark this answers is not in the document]
sbc100: > Apart form that, specifying the complete relocation type would expose relocation type names (which are for now an implementation detail of LLVM) to the wider text format.

The relocation type names are not indented to be LLVM specific.  The list of 20 relocation types, along with their ffull names, are listed above in this very document.

This is designed to mirror the ELF relocation types that are defined in the ELF header and not specific to either LLVM or GCC but are using in both place.

I think it might be a good idea to reflect this precisely in text for, so we can avoid having two different ways to specify things.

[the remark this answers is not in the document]
sbc100: > Personally I don't think `(@reloc ...)` or `(@sym ...)` should be elidable nor should there be a top-level annotation if each relocation/symbol is individually annotated. I think it'd be ok for `(@reloc $foo)` to automatically infer the kind of relocation depending on context, but I'd personally prefer that if something else were inserted that it's the full LLVM-derived name to avoid ambiguity. Like @sbc100 I'm not personally too worried about verbosity here. The assembly for `println!("hi!")` is way more verbose than the line of code, and IMO that's just kind of how assembly goes.

I'm OK with eliding relocations, if and only if we have a top level annotation.

I'd also be OK without a top level annotation if we can make all relocs explicit.

@alexcrichton, can I ask, what is your objection to the top level annotation?   Don't you think it would be nice to be able to, at a glace, distinguish relocation wat files without having to visually scan the whole wat file a @sym or @reloc?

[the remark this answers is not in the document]
alexcrichton: This might come down to how I'm approaching this from a different perspective (I think) than you might be. I'm thinking about this from a tooling perspective of a text-to-binary or binary-to-text transform. I suspect you're coming from the direction of producing-the-text and/or reading-the-text (please correct me though!). I'm concerned with "the binary should have one canonical text form and vice versa" and I'm not interested in auto-injecting `linking`/`reloc.*` sections myself (e.g. via some sort of CLI flag and/or configuration outside of the text file).

Not sure if that helps, but figured I'd write down how my perspective may be a bit different.

Linking.md
dschuff: This section should be updated to include the project that you're working on (either upstream in gcc, or a link to your fork), indicating that it supports both the binary and text formats, and that LLVM supports only the binary format. Also while you're at it, do you mind removing Binaryen from the list? That part is just incorrect, Binaryen has never supported this format.

feedab1e:
Hi, @dschuff proposed to move this PR forward. As far as I'm aware, the only issue left now is the tension between being explicit about relocations and compatibility with other metadata annotations. The crux of the issue is that, while interest exists for being able to discern at a glance whether the text given is relocatable or not, it cannot be done by requiring a relocation annotation to be present for each relocation entry.
The following example demonstrates this impossibility:
```wat
(module
  (func (@name "foo") (@sym)))
```
The issue here is that name annotations (as well as many others) create implicit references to WebAssembly entities which are not apparent in the text, so there is simply no place to attach a relocation annotation to that annotation, and since we require a relocation entry to be present for each reference to a WebAssembly entity, every program that uses such extensions would always fail object file validation. There are two ways of dealing with this issue:
1. tool-conventions has to keep track of and update every metadata annotation (even outside of tool-conventions) to be aware of linking metadata, and mandate explicit relocation annotations within those annotations, or
2. the linking format has to take into account the existence of other annotations, and allow generation of implied relocation entries to preserve compatibility.

Given the intrusivity of option 1, I find it impractical, and strongly advise against it and in favor of option 2.
As for differentiation for option 2, I think it would be adequate to just have a toplevel annotation as described in https://github.com/WebAssembly/tool-conventions/pull/258#issuecomment-3433334744.

@sbc100 @alexcrichton @dschuff, do you have any other comments?

sbc100:
> As for differentiation for option 2, I think it would be adequate to just have a toplevel annotation as described in [#258 (comment)](https://github.com/WebAssembly/tool-conventions/pull/258#issuecomment-3433334744).

I'm pretty strongly in favor of this top-level annotation.

Linking.md
```diff
@@ -734,3 +735,294 @@ necessary for referencing such segments (e.g. in `data.drop` or `memory.init`
...
+| `-<integer>` | Negative byte offset | `method` allows addend and `format` is signed |
+| `<labeluse>` | Byte offest to label | `method` is `*text`                           |
+
+- `symbol` describes the symbol against which to perform relocation.
+  - For `functext` relocation method, this is the function id, so that if the
+    addend is zero, the relocation points to the first instruction of that
+    function.
+  - For `customtext` relocation method, this is the name of the custom section,
+    so that if the addend is zero, the relocation points to the first byte of
+    data in that segment.
```
dschuff:
```suggestion
    data in that section.
```

feedab1e:
> > What happens to relocations in Type 1 sections if they reference symbols that are not included in the link?
> 
> They turn into tombstone values, the exact value of which depends on the type of section: https://github.com/llvm/llvm-project/blob/be955e5ac9a0b0c979221efb28b0f52aad7bd3d6/lld/wasm/InputChunks.cpp#L578-L595

Oh, wow! So, if in my custom section I have a relocation for a function entity that gets deleted, I just get the first function?

I myself would honestly expect relocations to mark the symbols live, so that never happens, and that's how I read the spec initially, but given this I'd like at least some documentation on that.

tgross35:
Gentle ping, are there any updates here? We have some interest in using these features for Rust's wasm inline assembly support.

feedab1e:
> Gentle ping, are there any updates here? We have some interest in using these features for Rust's wasm inline assembly support.

Not really, I'm going to take a crack at it again once the WASM backend lands in GCC.

Meanwhile, a partial implementation of this exists in my fork of WABT, linked in the initial comment.

tgross35:
> Not really, I'm going to take a crack at it again once the WASM backend lands in GCC.

Happen to have a link for this status? I've skimmed gcc-patches but didn't see too much.

feedab1e:
> > Not really, I'm going to take a crack at it again once the WASM backend lands in GCC.
> 
> Happen to have a link for this status? I've skimmed gcc-patches but didn't see too much.

https://gcc.gnu.org/wiki/WebAssemblyBackend, I plan to send the patches in the coming days.

daxpedda:
At `wasm-bindgen` we started using this in our [experimental next version](https://github.com/wasm-bindgen/js-bindgen). To that purpose @spxg has done a minimal implementation that only covers what we need: [`rwat`](https://github.com/Spxg/rwat).

Initially we can definitely say that stamping `@sym` and `@reloc` everywhere doesn't seem to be very productive. We do very much advocate for a global switch as has already been proposed a bunch of times, e.g. `(module (@relocatable))`.

Our next step is to make a full implementation of the proposed spec and hopefully push forward an experimental Rust implementation.

PR #262, opened 2025-12-19
Document that Wasm signatures support components.

sunfishcode:
The Wasm signature format described in Signatures.md is general enough to work on components, and there is [component support in wasmsign2], so document that the format supports components.

[component support in wasmsign2]: https://github.com/wasm-signatures/wasmsign2/pull/5

### DIFF
--- a/Signatures.md
+++ b/Signatures.md
@@ -1,6 +1,8 @@
-# WebAssembly module signatures
+# WebAssembly signatures
 
 This document describes a digital signature format specifically designed for WebAssembly modules.
+
+It also support [WebAssembly components](https://github.com/WebAssembly/component-model). Everything in this document that refers to "modules" also applies to "components".
 
 It satisfies the following requirements:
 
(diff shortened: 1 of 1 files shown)


jedisct1:
wasmsign2’s support for WASI components is untested, undocumented, and fairly rudimentary.

The core idea behind the signature format is to allow verification of subsets of a WebAssembly module, and to permit adding new sections without invalidating existing signatures. In this model, WebAssembly sections are the unit of verification.

I’m not very familiar with WASI components, but my assumption is that you would need the ability to split signatures by component instead of, or in addition to sections. Unless each component is represented as its own dedicated section, in which case the current scheme should work out of the box.

dgohman-fastly:
In Wasm components, [each component is represented as its own dedicated section](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md#component-definitions).

jedisct1:
Thanks!

Maybe something worth mentioning in the description.

sunfishcode:
I've now added a brief mention of how components are organized into sections.

PR #263, opened 2026-01-29
Update function annotation location

### LINKED ISSUE #251: Ambiguity in placement of metadata.code function annotations
CodeMetadata.md currently says that `@metadata.code.XXX` annotations must be within function definitions and gives an example:

```
(func (@metadata.code.hotness "\01") $test (type 0) ...)
```

But this could lead to ambiguity for annotations that could apply to either functions or instructions:

```
(func (@metadata.code.func_or_inst? "unknown") nop)
```

This potential ambiguity could be avoided (and the text format could be made nicer IMO) if the function annotations were placed before function definitions instead:

```
(@metadata.code.definitely_a_func "yep")
(func nop)
```

There are no formally specified function annotations, so this text format is only specified here in the tool-conventions repo. If we get stakeholders to agree, we should be able to change this.

cc @yuri91 as well as @fitzgen to route to appropriate wasm-tools folks, @sbc100 for WABT, and @kripken for binaryen. WDYT?

tlively:
As discussed in #251, function annotations should go before function definitions rather than after the `func` keyword. Update the example in CodeMetadata.md accordingly. Also update the example so that annotations consistently go on their own lines.

Fixes #251.

### DIFF
--- a/CodeMetadata.md
+++ b/CodeMetadata.md
@@ -60,13 +60,16 @@ Example:
 ```
 (module
   (type (;0;) (func (param i32 result i32)))
-  (func (@metadata.code.hotness "\01") $test (type 0)
-    (@metadata.code.branch_hint "\00") if
+  (@metadata.code.hotness "\01")
+  (func $test (type 0)
+    (@metadata.code.branch_hint "\00")
+    if
       i32.const 0
       local.set 0
     end
     local.get 1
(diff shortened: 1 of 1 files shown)


tlively:
cc @yuri91, @kripken

kripken (approved):

PR #268, opened 2026-05-23
Define a new "wasm-multivalue" calling convention

alexcrichton:
This commit is an attempt to define a new calling convention for WebAssembly which I've provisionally decided to call "wasm-multivalue". This new calling convention is inspired by discussions on #247 and #88, and the primary motivation of this calling convention is to be able to use multi-return signatures for intrinsics/functions in the component model.

This is a definition of a new, parallel, calling convention to the existing, now called "C", calling convention. The intention is that this avoids breakage to existing programs. The new calling convention is opt-in, requiring an annotation. The `multivalue` target feature enables usage of "wasm-multivalue", but enabling or disabling `multivalue`-the-target-feature has no effect on the "C" calling convention.

The technical definition of this new "wasm-multivalue" calling convention is to alter the previous calling convention in three ways:

* Primarily, `struct` returns are now returned directly if all fields are (optionally recursively) scalars. This means that returning a `struct`-by-value gets translated to multiple return values.

* To account for #88 this new calling convention additionally passes two-scalar-argument `struct` definitions directly instead of indirectly, matching what native calling conventions do for example.

* Finally, `__int128` and `long double` return values are now returned directly instead of indirectly.

On an implementation side of things, what I'd like to do is to gain consensus on this calling convention definition in the tool-conventions repository as a first step. Once the definition is settled I plan to go to LLVM/Clang to implement this calling convention, and after that I plan to go to the Rust compiler and implement the calling convention there. My plan for Rust is to implement `extern "wasm-multivalue" fn(..)` syntactically, and for C to use
`__attribute__((wasm_multivalue))` as the opt-in. Note that this is all bikesheddable, of course.

### DIFF
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -198,36 +198,80 @@ allows for a single result. The second calling is referred to as
 "wasm-multivalue" and assumes [multivalue] is enabled and additionally applies
 some small tweaks from the previous calling convention learned over time.
 
-The following table shows how all non-`struct` or `union` types are passed in
-both calling conventions:
+The following table shows how all types are passed in both calling conventions:
 
-Type                         | Parameter     | Result   |
------------------------------|---------------|----------|
-scalar[^1]                   | direct        | direct   |
-empty struct or union        | ignored       | ignored  |
-singleton struct or union[^2]| direct        | direct   |
-flat struct[^3]              | direct        | direct   |
-other struct or union[^4]    | indirect      | indirect |
-array                        | indirect      | N/A      |
+<table>
+  <tr>
+    <th>Type</type>
+    <th colspan="2">Parameter</th>
+    <th colspan="2">Result</th>
+  </tr>
+  <tr>
+    <th></th>
+    <th>"C"</th>
+    <th>"wasm-multivalue"</th>
+    <th>"C"</th>
+    <th>"wasm-multivalue"</th>
+  </tr>
+  <tr>
+    <td>8 to 64-bit scalars</td>
+    <td>direct</td>
+    <td>direct</td>
+    <td>direct</td>
+    <td>direct</td>
+  </tr>
+  <tr>
+    <td>128-bit scalars</td>
+    <td>direct</td>
+    <td>direct</td>
+    <td>indirect</td>
+    <td>direct</td>
+  </tr>
+  <tr>
+    <td>struct/union - 0 fields</td>
+    <td>ignored</td>
+    <td>ignored</td>
+    <td>ignored</td>
+    <td>ignored</td>
+  </tr>
+  <tr>
+    <td>struct/union - 1 scalar field</td>
+    <td>direct</td>
+    <td>direct</td>
+    <td>direct</td>
+    <td>direct</td>
+  </tr>
+  <tr>
+    <td>struct - 2 scalar fields</td>
+    <td>indirect</td>
+    <td>direct</td>
+    <td>indirect</td>
+    <td>direct</td>
+  </tr>
+  <tr>
+    <td>struct - 3+ scalar fields</td>
+    <td>indirect</td>
+    <td>indirect</td>
+    <td>indirect</td>
+    <td>direct</td>
+  </tr>
+  <tr>
+    <td>struct/union - other</td>
+    <td>indirect</td>
+    <td>indirect</td>
+    <td>indirect</td>
+    <td>indirect</td>
+  </tr>
+</table>
 
-[^1]: Signed 8 and 16-bit scalars are sign-extended, and unsigned 8 and 16-bit
-    scalars are zero-extended before passing or returning. `long double` and
-    `__int128` are passed directly as two `i64` values in paraemters, and in
-    results for the "C" convention they're returned indirectly and for the
-    "wasm-multivalue" calling convention they're returned directly.
+For scalars, signed 8 and 16-bit scalars are sign-extended, and unsigned 8 and
+16-bit scalars are zero-extended before passing or returning. Passing 128-bit
+scalars directly, such as `long double` and `__int128`, is done with two `i64`
+values.
 
-[^2]: Any struct or union that recursively (including through nested structs,
-    unions, and arrays) contains just a single scalar value and is not specified
-    to have greater than natural alignment.
-
-[^3]: This is only applicable to the "wasm-multivalue" ABI. Flat structs in
-    parameters are limited two structs with two fields which are themselves
-    (optionally recursively) scalars. A flat struct result requires that every
-    field of a struct is (optionally recursively) a scalar.
-
-[^4]: All other structs/unions not previously covered for both ABIs are
-    indirect. This is because "C" came before [multivalue], and for [multivalue]
-    not everything is represented as direct arguments.
+A "scalar field" of a struct or union is a field which recursively, through
+nested structs, unions, and arrays, contains just a single scalar vallue and is
+not specified to have greater than natural alignment.
 
 Some example C function signatures, and their corresponding WebAssembly function
 signatures, are:
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -7,7 +7,7 @@ compiler wishing to be ABI-compatible with it.
 
 The current version of this ABI is *1*.
 
-This ABI is designed to work with Release 1.0 of the WebAssembly [Specification](https://webassembly.github.io/spec/core/index.html). It does not require any 
+This ABI is designed to work with Release 1.0 of the WebAssembly [Specification](https://webassembly.github.io/spec/core/index.html). It does not require any
 [features](https://github.com/WebAssembly/proposals)
 that have not yet been implemented and standardized. Future versions will depend on
 features such as threads and/or multi-value.
@@ -16,7 +16,7 @@ features such as threads and/or multi-value.
 
 **Scalar Types**
 
-The ABI for the currently-specifed version of WebAssembly (also known as "wasm32") uses an "ILP32" data model, 
+The ABI for the currently-specifed version of WebAssembly (also known as "wasm32") uses an "ILP32" data model,
 where `int`, `long`, and pointer types are 32 bits. It is expected that the proposed extension to allow memories
 larger than 4GB ("wasm64") will use an "LP64" data model, where `int` is 32 bits while `pointer` and `long` are
 64 bits.
@@ -45,7 +45,7 @@ General type | C Type | `sizeof` | Alignment (bytes) | Wasm Value Type
  Floating point | `double` | 8 | 8 | f64
  Floating point | `long double` | 16 | 16 | (none)
  Floating point | `long double` (Emscripten) | 16 | 8 | (none)
- 
+
  * `long double` values correspond to 128-bit IEEE-754 quad-precision binary128 values.
  Operations on these values are currently implemented as calls to
  compiler-rt library functions.
@@ -122,31 +122,31 @@ to understand them.
 
 ## Locals and the stack frame
 WebAssembly does not have registers in the style of hardware architectures. Instead it has an
-unlimited number of function arguments and local variables, which have wasm value types. These 
+unlimited number of function arguments and local variables, which have wasm value types. These
 local values are generally used as registers would be in a traditional architecture, but there
 are some important differences. Because arguments are modeled explicitly and locals are local
 to a function, there is no need for a concept of callee- or caller-saved locals.
 
 
 ### The linear stack
-WebAssembly is a [Harvard](https://en.wikipedia.org/wiki/Harvard_architecture) architecture; 
+WebAssembly is a [Harvard](https://en.wikipedia.org/wiki/Harvard_architecture) architecture;
 this means that code and data are not in the same memory space. No code or code addresses are
 visible in the wasm linear memory space, the only "address" that a function has is its index
 in the wasm function index space. Additionally the wasm implementation's runtime call stack
 (including the return address and function arguments) is not visible in the linear memory either.
 This means that address-taken local C variables need to be on a separate stack in the linear memory
-(herafter called the "linear stack"). It also means that some functions may not need a frame in the 
+(herafter called the "linear stack"). It also means that some functions may not need a frame in the
 linear stack at all.
 
 Instead of registers visible to all functions, WebAssembly has a table of global variables. One of these
-acts as the stack pointer [TODO: describe how stack pointer is designated here, or in object file section] 
+acts as the stack pointer [TODO: describe how stack pointer is designated here, or in object file section]
 [TODO: discuss mutable global requirement].
 
 Each function may have a frame on the linear stack. This stack grows downward
 [TODO: describe how start position is determined].
-The stack pointer global (`SP`) points to the bottom of the stack frame and always has 16-byte alignment. 
-If there are dynamically-sized objects on the stack, a frame pointer (a local variable, `FP`) is used, 
-and it points to the bottom of the static-size objects (those whose sizes are known at compile time). 
+The stack pointer global (`SP`) points to the bottom of the stack frame and always has 16-byte alignment.
+If there are dynamically-sized objects on the stack, a frame pointer (a local variable, `FP`) is used,
+and it points to the bottom of the static-size objects (those whose sizes are known at compile time).
 If objects in the current frame require alignment greater than 16, then a base pointer (another local, `BP`)
 is used, which points to the bottom of the previous frame.
 The stack also has a "red zone" which extends 128 bytes below the stack pointer. If a function
@@ -164,8 +164,8 @@ Position                     | Contents                       | Frame
  `SP` + *d*<br>...<br>`SP`   | dynamic-size objects           | Current
  ...<br>`SP`-128             | small static-size objects      | Current (red zone)
 
-Note: in other ABIs the frame pointer typically points to a saved frame pointer (and return address) 
-at the top of the current frame. In this ABI the frame pointer points to the bottom of the current frame instead. 
+Note: in other ABIs the frame pointer typically points to a saved frame pointer (and return address)
+at the top of the current frame. In this ABI the frame pointer points to the bottom of the current frame instead.
 This is because the constant offset
 field of Wasm load and store instructions are unsigned; addresses of the form `FP` + *n* can be folded
 into a single insruction, e.g. `i32.load offset=n`. This is also why the stack grows downward (so `SP` + *n*
@@ -184,33 +184,107 @@ making a copy of any indirectly passed data that the callee should not modify.
 Similarly, types can either be returned directly from WebAssembly functions or
 returned indirectly via a pointer parameter prepended to the parameter list.
 
-Type                         | Parameter     | Result   |
------------------------------|---------------|----------|
-scalar[1]                    | direct        | direct   |
-empty struct or union        | ignored       | ignored  |
-singleton struct or union[2] | direct        | direct   |
-other struct or union[3]     | indirect      | indirect |
-array                        | indirect      | N/A      |
-
-[1] `long double` and `__int128` are passed directly as two `i64` values.
-Signed 8 and 16-bit scalars are sign-extended, and unsigned 8 and 16-bit
-scalars are zero-extended before passing or returning.
-
-[2] Any struct or union that recursively (including through nested structs,
-unions, and arrays) contains just a single scalar value and is not specified to
-have greater than natural alignment.
-
-[3] This calling convention was defined before
-[multivalue](https://github.com/WebAssembly/multi-value) was standardized. A new
-default calling convention that changes this behavior and takes advantage of
-multivalue may be introduced in the future.
-
 Varargs are placed in a buffer by the caller and the last parameter to the
 function is a pointer to that buffer. The callee is allowed to modify the
 contents of the buffer, so the caller is responsible for making a copy of any
 varargs data that the callee should not modify. Arguments are arranged in order
 in the buffer, with each argument placed at the lowest appropriately aligned
 address after the previous argument.
+
+There are two calling conventions defined for use at this time. The default
+calling convention is typically referred to as the "C" calling convention and
+was created before [multivalue] was available, thus only
+allows for a single result. The second calling is referred to as
+"wasm-multivalue" and assumes [multivalue] is enabled and additionally applies
+some small tweaks from the previous calling convention learned over time.
+
+The following table shows how all non-`struct` or `union` types are passed in
+both calling conventions:
+
+Type                         | Parameter     | Result   |
+-----------------------------|---------------|----------|
+scalar[^1]                   | direct        | direct   |
+empty struct or union        | ignored       | ignored  |
+singleton struct or union[^2]| direct        | direct   |
+flat struct[^3]              | direct        | direct   |
+other struct or union[^4]    | indirect      | indirect |
+array                        | indirect      | N/A      |
+
+[^1]: Signed 8 and 16-bit scalars are sign-extended, and unsigned 8 and 16-bit
+    scalars are zero-extended before passing or returning. `long double` and
+    `__int128` are passed directly as two `i64` values in paraemters, and in
+    results for the "C" convention they're returned indirectly and for the
+    "wasm-multivalue" calling convention they're returned directly.
+
+[^2]: Any struct or union that recursively (including through nested structs,
+    unions, and arrays) contains just a single scalar value and is not specified
+    to have greater than natural alignment.
+
+[^3]: This is only applicable to the "wasm-multivalue" ABI. Flat structs in
+    parameters are limited two structs with two fields which are themselves
+    (optionally recursively) scalars. A flat struct result requires that every
+    field of a struct is (optionally recursively) a scalar.
+
+[^4]: All other structs/unions not previously covered for both ABIs are
+    indirect. This is because "C" came before [multivalue], and for [multivalue]
+    not everything is represented as direct arguments.
+
+Some example C function signatures, and their corresponding WebAssembly function
+signatures, are:
+
+```c
+
+// [] -> []
+void f1(void);
+
+// [f32 f64] -> [i32]
+int f2(float, double);
+
+// [i32 i64 i64] -> []
+//    |   |   ^ high bits of parameter
+//    |   ^ low bits of parameter
+//    ^ return-pointer
+__int128 f3(long double);
+
+// [] -> []
+struct Foo4 { }
+union Bar4 { }
+Bar4 f4(Foo4);
+
+// [i32] -> [i32]
+struct Foo5 { int a; }
+union Bar5 { int a; }
+Bar5 f5(Foo5);
+
+// [i32 i32] -> []
+//    |   ^ pointer to parameter
+//    ^ return-pointer
+struct Foo6 { int a; int b; }
+Foo6 f6(Foo6);
+
+// [i32 i32] -> [i32 i32]
+struct Foo7 { int a; int b; }
+__wasm_multivalue__
+Foo7 f7(Foo7);
+
+// [] -> [i64 i64]
+__wasm_multivalue__
+__int128 f8(void);
+```
+
+Note that the "C" ABI is the default ABI used in C, and using the
+"wasm-multivalue" ABI requires an explicit opt-in. While the default calling
+convention may change in the future that is not yet planned at this time.
+Additionally note that programs can use either ABI to define functions, but the
+caller and callee calling convention must match when a call is done.
+
+To use the "wasm-multivalue" calling convention the [`multivalue` target
+feature](./Linking.md#target-features-section) must be enabled. It is a complier
+error to use this calling convention when the target feature is not enabled.
+Additionally note that enabling the `multivalue` target feature does not change
+the "C" calling convention, it only enables usage of "wasm-multivalue".
+
+[multivalue]: https://github.com/WebAssembly/multi-value
 
 ## Program startup
 
--- a/BasicCABI.md
+++ b/BasicCABI.md
@@ -249,7 +249,7 @@ The following table shows how all types are passed in both calling conventions:
     <td>direct</td>
   </tr>
   <tr>
-    <td>struct - 3+ scalar fields</td>
+    <td>struct - 3-100 scalar fields</td>
     <td>indirect</td>
     <td>indirect</td>
     <td>indirect</td>
@@ -270,7 +270,7 @@ scalars directly, such as `long double` and `__int128`, is done with two `i64`
 values.
 
 A "scalar field" of a struct or union is a field which recursively, through
-nested structs, unions, and arrays, contains just a single scalar vallue and is
+nested structs, unions, and arrays, contains just a single scalar value and is
 not specified to have greater than natural alignment.
 
 Some example C function signatures, and their corresponding WebAssembly function


tlively:
SGTM! I like that making this opt-in lets us avoid the question of how large a struct should be allowed to be returned by value.

BasicCABI.md
```diff
@@ -184,34 +184,108 @@ making a copy of any indirectly passed data that the callee should not modify.
...
+scalar[^1]                   | direct        | direct   |
+empty struct or union        | ignored       | ignored  |
+singleton struct or union[^2]| direct        | direct   |
+flat struct[^3]              | direct        | direct   |
+other struct or union[^4]    | indirect      | indirect |
+array                        | indirect      | N/A      |
+
+[^1]: Signed 8 and 16-bit scalars are sign-extended, and unsigned 8 and 16-bit
+    scalars are zero-extended before passing or returning. `long double` and
+    `__int128` are passed directly as two `i64` values in paraemters, and in
```
fitzgen:
```suggestion
    `__int128` are passed directly as two `i64` values in parameters, and in
```

BasicCABI.md
```diff
@@ -184,34 +184,108 @@ making a copy of any indirectly passed data that the callee should not modify.
...
+both calling conventions:
+
+Type                         | Parameter     | Result   |
+-----------------------------|---------------|----------|
+scalar[^1]                   | direct        | direct   |
+empty struct or union        | ignored       | ignored  |
+singleton struct or union[^2]| direct        | direct   |
+flat struct[^3]              | direct        | direct   |
+other struct or union[^4]    | indirect      | indirect |
+array                        | indirect      | N/A      |
```
fitzgen: FWIW, I think this would be more readable as either two tables (one for "c" and one for "wasm-multivalue") or a nested/multi-dimensional table, rather than using footnotes here.

Something like this:

<table>
  <tr>
    <th>Type</type>
    <th colspan="2">Parameter</th>
    <th colspan="2">Result</th>
  </tr>
  <tr>
    <th></th>
    <th>"C"</th>
    <th>"wasm-multivalue"</th>
    <th>"C"</th>
    <th>"wasm-multivalue"</th>
  </tr>
  <tr>
    <th>scalar</th>
    <td>direct</td>
    <td>direct</td>
    <td>direct</td>
    <td>direct</td>
  </tr>
  <tr>
    <th>emptry struct or union</th>
    <td>direct</td>
    <td>direct</td>
    <td>direct</td>
    <td>direct</td>
  </tr>
  <tr>
    <th>singleton struct or union</th>
    <td>direct</td>
    <td>direct</td>
    <td>direct</td>
    <td>direct</td>
  </tr>
  <tr>
    <th>flat struct or union</th>
    <td>indirect</td>
    <td>direct</td>
    <td>indirect</td>
    <td>direct</td>
  </tr>
  <tr><td colspan="5">etc...</td></tr>
</table>

As-is, it currently makes it look like the two calling conventions don't actually differ from each other at a glance.

alexcrichton: Good suggestion! Updated

alexcrichton:
I've got a concrete implementation of this at https://github.com/llvm/llvm-project/pull/200076, although I'm by no means an LLVM expert so that's unlikely its final form. It's at least the general thrust, however.

dschuff:
This sounds good to me too. But I don't think we can/should avoid the question of picking concrete size though, right? Once this starts being used by component model intrinsics, we'll want it to be stable, no?

dschuff:
Oh, I guess @tlively you're saying that the convention will be to just return everything byval, and just never make this the default, and then the opt-in would have to be done by the programmer or whoever defines an API?
It seems like it might be nice to have it as the default in the future though if it's really beneficial.

One thing that we could do: The [GlobalOpt](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/IPO/GlobalOpt.cpp#L2026) pass knows how to change functions to use the `FastCC` calling convention when they are private enough. Maybe we can make FastCC an alias for this, or make a TTI hook to choose which CC to upgrade to instead of just using FastCC. But that might only make sense if the calling convention was stable and known to be almost always beneficial. 
(Or we could have some more sophisticated analysis in GlobalOpt to decide whether it's beneficial... or just have FastCC and wasm-multivalue CC be the same except for the number of return values used... anyway there are several options, but we wouldn't need to do anything fancy if wasm-multivalue were stable and mostly an improvement).

alexcrichton:
@dschuff to confirm you mean a limit to the number of return values, right? If so, I was wondering the same, but testing locally it looked like there was no limit to the number of parameters that would be generated so I figured I'd follow the same. I think it'd be reasonable to pick a limit, however, and start spilling to the stack afterwards. Implementing that scale of ABI change is probably going a bit beyond my LLVM abilities personally, but otherwise, yes, we'll want to consider this stable in relatively short-order to be able to use it in the component model.

I do agree that it'd probably be worthwhile to make this the default ABI one day (or at least a variant thereof). I also agree that it'd be reasonable to switch internal functions to this ABI automatically, and this is something I plan on measuring for Rust at some point and probably applying (the precise ABI of `fn()` is perma-unstable unlike `extern "C" fn()` in Rust, exactly to be able to make changes like this).

For now though my plan is to start relatively conservative and mostly just unblock the ability to use multi-return in wasm modules (and subsequently component model intrinsics/lowerings/etc). I don't have immediate plans personally to investigate optimization passes or plan out a switch for this to become the default ABI.

fitzgen:
I think at minimum the limit would need to match the web spec’s implementation limits, no?

alexcrichton:
Testing locally, currently there's no limit in LLVM to match the web 1000 parameter limit. IIRC historically LLVM can also emit more locals than the web allows, so I'm not sure how many preexisting limits are adhered to within LLVM.  

If a limit were to be added, though, it'd probably be something like ~400 ABI-level parmeters and ~400 field structs. Half of the 1000 limit due to some types becoming two ABI values (e.g. `i128`) and a little less to account for things like return poitners and such being injected. 

I'm happy to write this down, but I'm also a bit wary of diverging too much from the preexisting ABI. In practice I suspect we can retroactively apply limits at any time since 400 parameters is a lot and this isn't super widespread yet, too.

tlively:
> Oh, I guess @tlively you're saying that the convention will be to just return everything byval, and just never make this the default, and then the opt-in would have to be done by the programmer or whoever defines an API?
> It seems like it might be nice to have it as the default in the future though if it's really beneficial.

Right. Having it be opt-in completely avoids the question of how many return values to support before spilling. I agree that it would be nice to have a default-on or FastCC multivalue ABI that only uses a reasonable, finite number of return values (2? 4? 8?), but by starting with something opt-in we can make incremental progress while continuing to defer a decision on that to the future.

@alexcrichton, another consideration is that there are several known bugs with the multivalue implementation at the moment (e.g. https://github.com/llvm/llvm-project/issues/98323, https://github.com/llvm/llvm-project/issues/92995). The last status here was that @sunfishcode was looking into some of them.

dschuff:
Yeah, previously we had discussed having an ABI that was meant to be similar to existing machine ABIs that can pass and return a limited number of elements in registers, for a performance boost compared to using the stack. The idea was that this could hopefully be implemented by also using registers for at least most of those parameters/return values, but that beyond a small number (probably less than 10, as Thomas mentioned), keeping parameters that would certainly need to be in memory as explicitly in memory made sense. The reason there is currently no limit (on the number of parameters used to pass struct elements, or the number of return values) is just that nobody has done the performance measurement work to figure out what the best number is.

It sounds like you are looking for something a little different though, i.e. that a potentially-huge number of params or returns is what you might actually want? As Thomas says, this has the nice property that nobody has to do that performance measurement (and that performance maybe isn't really the motivation for your use case anyway?).

I think I would really like us to actually just do some experiments though, and pick a small number to use as default. As I mentioned, I think it can unlock some easy performance wins for LTO and potentially as a default. We hadn't really considered the possibility of a large or unlimited number of params. It's certainly a simpler ABI, but I would want to look at the consequences (in terms of code size, generated code size, and performance) of that. I could imagine a code size blowup (pushing 100 values onto the stack at the callsites vs just passing a pointer), a performance cliff (2x copies for all those parameters? hitting some pathological case in the register allocator?); or potentially even a performance gain (having 100 values on the wasm stack instead of linear memory means no bounds checks in wasm64 mode). But we don't know.

dschuff:
> I'm happy to write this down, but I'm also a bit wary of diverging too much from the preexisting ABI. In practice I suspect we can retroactively apply limits at any time since 400 parameters is a lot and this isn't super widespread yet, too.

Do you mean the preexisting mult-value ABI? Don't worry about divergence from that, it's not used anywhere and as Thomas mentioned, it doesn't even work yet.

QuantumSegfault:
What's the rationale for limiting the parameter aggregate split to only two scalars? I understand there should be a limit, but I imagine a higher limit, say 4, would allow for better utilization of the new ABI (the example in mind is 3D math & graphics; vec3, vec4, quat, color).

dschuff:
If that were the limit (it isn't yet), it would be because we found empirically that it performed better. It would be interesting to know for example, how many parameters (if any?) can be passed in registers in wasm implementations and whether the builtin spilling is any worse than passing in wasm memory would be.
But again, nobody has done this experimentation yet.  I agree it seems likely that we'd want at least 4 though.

dschuff:
Oh, sorry I missed that the current proposal and PR does actually use exactly 2 scalar arguments. I think what I said above about our old discussions still applies; I think it might make sense to have more than 2, based on performance and/or your usability argument. @alexcrichton may have just picked it based on the component models' needs, I would actually be curious if there was more to it than that.

alexcrichton:
@tlively oh thanks for pointing out those issues! Are you or others aware of more issues than those? I can try to dig into them and see if I can't figure them out.

Otherwise though, one point worth clarifying here is that my intention is to implement this ABI in both Rust and Clang at which point the definitions need to match, so that'll at least place a constraint on the degree of flexibility we have for tweaking the ABI over time. My personal intention is to unblock the ability to use multi-value in the component model where the only requirement is some way to write a function in C/Rust that maps to a specific wasm type signature, but it's all generated code so is relatively easy to adjust/modify compared to a bunch of handwritten code in the ether.

Also to your points @dschuff, sorry no I don't mean to give the impression that I have a use case with tons of params/results, that was mostly in response to @fitzgen's thought that LLVM may want to match the JS embedder API limits for wasm which places a hard requirement that functions can't have more than 1000 parameters or 1000 results. Additionally by matching the previous ABI I mean the default one day (called "C" after this PR) which, from what I can tell, has no limit on the number of parameters.

---

w.r.t. speed/optimization, I think it's going to be a bit of a tricky issue. Number-of-parameters-in-registers varies across architectures/OSes which means that a benchmark on x64 linux will probably have different results than aarch64 macos will probably have different results than x64 Windows. My main goal here is to expose the ability to use multi-value returns, or effectively be able to map any wasm function type signature to some possible C/Rust function type signature, and in that sense I'd prefer to not get too bogged down in perf investigations. At the same time though I'd predict not a whole lot of ability to tweak this ABI after-the-fact, so I'd ideally like to make sure it's "right" and takes all concerns into account.

> The reason there is currently no limit (on the number of parameters used to pass struct elements, or the number of return values) is just that nobody has done the performance measurement work to figure out what the best number is.

To ask a question about this: this isn't something that can easily be retroactively modified though, could it? That would amount to changing the default ABI, which I imagine has all sorts of thorny issues because it's a "break the world" moment where nothing before can be mixed with anything after.

> I could imagine a code size blowup, a performance cliff; or potentially even a performance gain

For this point in particular, on one hand my goal with this new ABI is to be able to map any wasm function type signature to a C/Rust type signature. If a limit were applied to parameters or results then I'd argue it should be high enough that nothing would reasonably want to use or generate such a wasm type signature, but this limit would probably be well above the threshold of "useful for performance".

For example a semi-plausible situation is that in the component model the ABI for a particular imported function ends up using 16 return values. I don't think any native platform reasonably has support for 16 return registers, but from a code generation perspective it's still something that'd be desirable to be able to generate code for. This could of course inform the other way, though, by having stricter limits in the component model (e.g. at most 4 returns or something ilke that). So far we don't have a definition of what a multi-return ABI would be for the component model, but what we do have in the meantime is a set of functions that would be desirable to return up to 3 values (at least thinking about intrinsics off the top of my head).

Put another way, this ABI, while I think it's an improvement over the prior,  may not be best served trying to be the end-all-be-all. To me there's probably 3 desirable ABIs for WebAssembly right now:

1. The current "C" which is the default. Mostly around for historical compatibility at this point.
2. Some ABI which is suitable for mapping any reasonable wasm function type signature into C/Rust, and that's what "wasm-multivalue" is being used for in this PR.
3. Some ABI tuned for the best performance, placing more strict limits on params/results to solve some of the hypothetical problems you mention.

ABIs are always a tricky thing...

> Do you mean the preexisting mult-value ABI? Don't worry about divergence from that

Ah, yes, indeed! I meant following in the footsteps of the "C" ABI today (with no limit on params). I agree there's no need to strictly follow the preexisting multivalue abi.

---

@QuantumSegfault 

> What's the rationale for limiting the parameter aggregate split to only two scalars?

Honestly not a ton of rationale. I tried some godbolt code and saw that on x86_64 the limit was 2 scalars being flattened to registers and decided to mirror that. I'm not opposed to upping the max. One caution though is that parameters to wasm functions are typically mapped to native ABIs/registers at the native-machine-code level of which there really aren't all that many. Flattening a 4-element vector would spill all other arguments to the native machine stack on x86_64 for example in Wasmtime (I think at least), meaning that wasm-level parameters are still memory-level parameters when executed. In that sense maximizing splitting or more aggressively splitting won't necessarily lead to more performance. I'd guess it'd help clean up the wasm code slightly, but the complications would mostly jsut be shifted to native.

---

> It would be interesting to know for example, how many parameters (if any?) can be passed in registers in wasm implementations

I can speak to this from the Wasmtime side of things at least for integer arguments/results in registers:

| platform | argument registers | return registers |
|-----------|-----------|-------------------|
| x86_64  | 4 | 8 |
| aarch64 | 6 | 6 |
| riscv64 | 8 | 2 |
| s390x | 4 | 6 | 

(wow we're kind of all over the place...)

That being said, for Wasmtime we have tight control over the ABI (e.g. we're not trying to match system ABIs) so we can adjust these and tweak them.

One reason I'm hesitant to place too many limits in wasm is that if someone is trying to optimize for one particular engine or one particular architecture in one engine I'd want to avoid closing off possibilities by having a least-common-denominator of sorts or something like that. That's where I'd imagine that the limits for params/results (and maybe flattening?) would be relatively high to be beyond what any native architecture can reasonably implement. As I say that out loud though, on the performance part, it's probably bad to spill to both the wasm stack and then also spill to the native stack. It's probably best to spill to one or the other, and in that situation it might make the most sense to pretty aggressively use params/results in wasm to enable the most possible optimizations in engines...

---

apologies for the lots-of-words, but I definitely want to acknowledge that this is not an easy problem nor something I was planning on landing super fast. Thank you everyone for chiming in on this, I very much appreciate it!

dschuff:
> > The reason there is currently no limit (on the number of parameters used to pass struct elements, or the number of return values) is just that nobody has done the performance measurement work to figure out what the best number is.
> 
> To ask a question about this: this isn't something that can easily be retroactively modified though, could it? That would amount to changing the default ABI, which I imagine has all sorts of thorny issues because it's a "break the world" moment where nothing before can be mixed with anything after.

Sorry I think we may be talking about different things. I meant that (I think) in the current experimental-mv there's no limit on the number of struct elements (whereas the default C ABI doesn't pass any). We can change it now, since there are no users. But yeah obviously anything that becomes a default now or in the future can't easily be changed once it has.

One thing I wasn't thinking about yet is that there currently is no limit on the number of wasm argument parameters even in the default ABI. But this is less of an issue because there is just one wasm param for each C param, and there just aren't very many C functions with 1000 parameters. So nobody has complained about hitting the implementation limit. If we start destructuring structs into wasm params (or arrays in structures?) then it becomes much more likely to be an issue.

I do think it can make sense to treat the implementation incrementally though. By having an explicit opt-in less-limited (unlimited?)-destructuring ABI, we can solve a subset of the problems (plumbing new calling conventions through the layers, mixing conventions, working out LTO bugs, etc) without needing to pin down a number. We will still have to figure out things like arrays-in-structures, counting float vs int separately, and probably other stuff I'm forgetting, some of which is probably already handled in experimental-mv. 
If destructuring is limited (like your proposed limit of 2 for args), there's also the question of whether we want to skip such structs completely (passing them by ref) or destructure the first 2 and pass the rest by reference (writing this out, the former sounds better, actually...).


Gemini tells me (haven't verified) that V8 uses 5 argument registers and 2 return registers on x86-64 and 6 arg / 2 return on ARM64. So an interestingly large difference in the number of return registers.
Raw intuition suggest to me that passing struct args might be more common or have a bigger impact on performance than returning structs, but that's based on no actual evidence.

alexcrichton:
That's a good point yeah, returning a giant (possibly generated) struct seens more likely than passing 1000 params. For incrementality one point of leverage there is that Rust has explicit stability for all ABIs, meaning that the new `extern "wasm-mutlivalue" fn()`, when added, will be unstable and nightly-only. That means that even after landing this and adding this to Rust there'll be at least a 6 week window to shake out bugs, adjust the definition, etc, without impacting users in the wild. Basically I think it's reasonable to expect that this isn't insta-stable upon landing in LLVM.

For argument destructuring my loose intention was to resolve https://github.com/WebAssembly/tool-conventions/issues/88 while additionally following the conventions of native ABIs. I think a big difference between wasm and native though is that wasm has infinite "registers" and native doesn't, meaning that the destructuring just keeps happening in wasm where it eventually is guaranteed to stop on native. I do think it would be useful to pass at least some arguments destructured (e.g. a Rust `&str` is a pointer/len combo), but I also agree that only destructing N is a bit odd.

How about this maybe:

* Params stay as-is in this PR -- unlimited amount and 2-field structs unconditionally destructured. In practice placing a cap on this later seems unlikely to break anyone in practice due to the general convention that things have <20 arguments.
* Result destructuring is capped at structs of 100 fields. Should still be possible to write a C function for any reasonable wasm signature while also ensuring giant structs don't immediately become invalid wasms when returned.

tlively:
> @tlively oh thanks for pointing out those issues! Are you or others aware of more issues than those? I can try to dig into them and see if I can't figure them out

The issues from this search look relevant: https://github.com/llvm/llvm-project/issues?q=is%3Aissue%20state%3Aopen%20label%3Abackend%3AWebAssembly%20experimental-mv. Searching for "multivalue" instead of "experimental-mv" gives more results, but I don't think they're all relevant.

alexcrichton:
Thanks! Going through all those, I originally expected some large work needing to be necessary to make this robust but I actually think things are in a pretty good spot. With a few minor fixes I think that all that's really needed to ship multivalue is a definition of what it actually does. Some notes from those issues:

* https://github.com/llvm/llvm-project/issues/70402 - fixed by https://github.com/llvm/llvm-project/pull/200514
* https://github.com/llvm/llvm-project/issues/153567 - fixed by https://github.com/llvm/llvm-project/pull/200514
* https://github.com/llvm/llvm-project/issues/133292 - this is an issue about (1) forgetting to specify `experimental-mv` across an LLVM upgrade boundary, and (2) that the backend should probably emit an error for `musttail` functions that aren't turned into tail calls. Not a problem for the ABI definition here, however.
* https://github.com/llvm/llvm-project/issues/92995 - the prevailing assumption so far is that `experimental-mv` is missing somewhere in this LTO pipeline. Without a reproduction I'm not sure how to investigate much further. I noted in the issue though that the design here, with a first-class ABI, I think will make operations like LTO more robust by default since the ABI will be threaded without needing a flag somewhere.
* https://github.com/llvm/llvm-project/issues/146528 - this a consequence of the loose definition of the preexisting `experimental-mv` ABI where all records are returned in "direct". What exactly "direct" mode does for various types is, I believe, expected to be surprising. The `wasm-multivalue` ABI proposed here, and my prototype implementation in Clang for example, does not attempt to pass this structure by-value and it remains in-memory. Effectively I believe that the checks and limitations on what to flatten specified here are necessary for the ABI to work as expected. Regardless, this issue doesn't currently affect the `wasm-multivalue` ABI.
* https://github.com/llvm/llvm-project/issues/71124 - my guess is that this may be a consequence https://github.com/llvm/llvm-project/issues/98323 or a consequence of not compiling everything in the world with `experimental-mv`. I'm a bit hesitant to download random binaries myself from colorful domains and try to reproduce, so I haven't dug further.
* https://github.com/llvm/llvm-project/issues/136506 - a real issue, but IMO not a blocker as it's debuginfo-related

Overall it seems like after https://github.com/llvm/llvm-project/pull/200514 and https://github.com/llvm/llvm-project/pull/200429 there's no at least documented issue with a hypothetical new `wasm_multivalue` ABI. Not to say it's bulletproof, but a better starting point than I expected!

alexcrichton:
I've now updated this with a 100-field-limit on returned `struct`s to have at least some limit there. I've additionally done the Rust side of things over at https://github.com/rust-lang/rust/pull/157334.

eisenwave:
If I understand it correctly, then something like
```cpp
struct rgb { u8 r, g, b; };
```
would be passed indirectly because it contains three scalar fields, or is it about the three `u8`s collectively occupying a single scalar field, so it does get passed directly?

This kind of thing gets passed by single register in the x86_64 psABI, and developers have come to rely on the idea that they can equally pass a 32-bit `struct` with four member as well as a 32-bit integer, and both will be equally good in the ABI.

I don't want to go back to caveman technology like Java developers who pass RGB colors as `int` because creating an object is too expensive, so I'd be strongly opposed to an ABI that does not pass `struct rgb` as a single `i32`.

eisenwave:
And I'm not even sure if you thought of the `struct rgb` case. None of the examples involve sizes less than `i32` currently, so it's not very noticeable how punishing the ABI is toward using `array<u8, 4>` instead of `u32`.

> Honestly not a ton of rationale. I tried some godbolt code and saw that on x86_64 the limit was 2 scalars being flattened to registers and decided to mirror that.

Well yeah, but the x86_64 psABI has a limit of 2x64-bit rather than a limit of 2xAnything, and I'm sure a lot of thought went into designing that ABI in such a way. Your ABI is vaguely inspired by x86_64 but also has a dramatically different system of adding up scalars, and I don't want performance characteristics of the native ABI to be dramatically different from the WASM ABI.

fitzgen:
Packing multiple 8- and 16-bit integers into a single `i32` parameter implies additional Wasm instructions in every function prologue to unpack them and, even worse, for every call site to do the packing. That is pretty bad for code size and likely won't be boiled away by the Wasm engine's optimizer either.

Passing via `i32` (with unspecified upper bits to avoid forcing sign/zero extend instructions) is a much better trade off. Especially since Wasm doesn't have a narrower-than-32-bits value type so those integers will be `i32` values on the operand stack already anyways.

From a Wasm engine's optimizer's perspective, the most important thing is to avoid spilling operand stack values to memory (that is, to avoid passing arguments and returns indirectly via pointers as much as possible/reasonable) because Wasm's semantics make it very difficult to elide memory operations (especially stores). Even if a call was inlined, eliding the writes into the shadow stack memory that was the call's return pointer area is quite difficult and unlikely to happen in practice.

eisenwave:
> Packing multiple 8- and 16-bit integers into a single `i32` parameter implies additional Wasm instructions in every function prologue to unpack them and, even worse, for every call site to do the packing. That is pretty bad for code size and likely won't be boiled away by the Wasm engine's optimizer either.

At worst, the packing results in a bit of bit manipulation, and that is generally not too expensive. It doesn't always get optimized away, and that is a known tradeoffs that native ABIs have taken when they decided to do bit packing.

To be fair, there is a lot more register pressure in native ABIs, so you need to bit-pack to not run out of registers too fast. Despite us not really having that problem, we still spill into memory much more.

> Passing via `i32` (with unspecified upper bits to avoid forcing sign/zero extend instructions) is a much better trade off. Especially since Wasm doesn't have a narrower-than-32-bits value type so those integers will be `i32` values on the operand stack already anyways.

Yeah, I like passing by `i32` automatically. I don't want to manually rewrite my code so that `structs` containing three bytes, containing a status code and two 16-bit indices, etc. are all passed expensively despite developers generally understanding that you should pass small structs by value since that is cheap.

> From a Wasm engine's optimizer's perspective, the most important thing is to avoid spilling operand stack values to memory (that is, to avoid passing arguments and returns indirectly via pointers as much as possible/reasonable) because Wasm's semantics make it very difficult to elide memory operations (especially stores). Even if a call was inlined, eliding the writes into the shadow stack memory that was the call's return pointer area is quite difficult and unlikely to happen in practice.

Which is exactly why you want that packing to happen. The limit is currently set at two scalars, meaning that every `struct rgb` and `struct argb` of 3 or 4 bytes is getting spilled into memory, whereas it gets passed via single 32-bit register in the x86_64 psABI.

These rules seem dramatically different to me, and they necessitate changing your programming style and abandoning zero-overhead abstractions just for ABI purposes. The general consensus in C++ is that if a `struct` is two pointers in size or less, it's cheap to pass by value, and you should do so. Lots of code has been written with that assumption already. Those teachings can be abandoned if the backend is wasm-multivalue because the rule for the first time in 40 years becomes "if a `struct` has two data members or fewer".

alexcrichton:
On one hand my main motivation here is to expose multiple returns through a calling convention, and I don't want to get too bogged down in an attempt to create the perfect ABI such that this never ends up getting finished. On the other hand I understand that this is the ideal time to do such ABI construction/crafting and it'll only be harder to change in the future. To that extent I think it's worth considering what you're mentioning @eisenwave, but I'm personally still trying to be cognizant of "I don't want to bikeshed this ABI to death" such that the end state is we still don't have a multi-return calling-convention.

@eisenwave you're right that I was originally testing with aggregates-of-64-bit-values hence the current conclusion of "matching native ABIs". The rules for x86_64 seem to basically say that if a structure fits in 16-bytes it's bit-packed in two registers and passed as an aggregate. For an ABI detail that seems reasonable to me as the bit-packing is probably much more efficient than spilling to/from memory, and if it's what's done on native that's probably good enough for wasm. I compared a bit with aarch64 as well and it seems to roughly agree with x86_64. I'll note though that the comparison with native can only go but so far because wasm is architecture independent and doesn't know how many registers nor what width they have of the underyling native platform.

@fitzgen I don't fully agree with your overall conclusion that the bit-packing here is going to be more expensive than it's worth, but I also agree with you to an extent too. For the use case of rgb with three 1-byte values the current ABI would specify that wasm loads/stores would be used to pass the struct-by-value since it's got 3 fields and thus is required to be indirect. With the bit-packing this can be passed in a single `i32` value, in theory. For this specific case I suspect that bit-packing the three 8-bit values into one 32-bit value is probably more efficient than passing through memory, but I mostly say that because it's what native does in this case, and I'd naively assume someone probably did actual benchmarking (as opposed to me doing none) on native. Basically, given the prior art of native, it seems reasonable to consider this as a more efficient alternative than passing things through memory.

As to what this might look like for wasm, I'm not entirely sure. On x86_64 there's rules for mixing floats/ints, so for example if you [pass a float/int together](https://godbolt.org/z/1x199hKq8) they're bit-packed in one register. Unless anyone feels strongly, I'd prefer to avoid that as it seems like a bit too much complexity for wasm. Wasm only has `i32` and `i64` types otherwise and additionally has "infinite registers". This means that a wasm runtime could translate two `i32` arguments to a function to a single native 64-bit integer argument (bit-packed) to save on registers. Given all that, what I'd propose is probably:

* For structs 1-4 bytes in size made up of only integers, they're passed as a bit-packed i32
* For structs 5-8 bytes in size made up of only integers, they're passed as a bit-packed i64
* For structs 9-12 bytes in size made up of only integers, they're passed as a bit-packed i64 and i32
* For structs 13-16 bytes in size made up of only integers, they're passed as two bit-packed i64s
* For all other structs, the preexisting rules in this PR apply.

This means that small structs-of-integers are optimized similarly to native while small structs of integers-or-floats are optimized the same way as this PR proposes today. How's that sound?

eisenwave:
> This means that small structs-of-integers are optimized similarly to native while small structs of integers-or-floats are optimized the same way as this PR proposes today. How's that sound?

Those rules seem pretty fine. It may be worth considering what happens when you pack `struct rgb` and `float` into one `struct`. I would find it surprising passing those separately was `i32 + f32`, but when wrapping them in a `struct`, it would cause indirect passing of `4xi32 + f32` spilled memory.

So you would probably want that to be passed as `i32 + f32` based on the rule that you bit-pack integers and separate floats. I'm just not sure what rules get you there. The x86_64 psABI seems to pack floats into the low 64 bits of xmm registers while leaving the upper 64 bits unused, which is just weird.

In general, the floating-point part may deserve a bit more thought. Consider that an extremely common aggregate type in linear algebra is `vec3` or `vec4`, i.e. 3-4 floats passed at once. You use this for 3D/4D positions, ARGB floating-point colors, etc. etc. etc. So I would suggest to have a separate pool for integers and for floats, and pass up to four floats separately rather than indirectly.

To be clear, I'm not suggesting that you should pack `vec2f32` into a single `f64` or something like that, just pass floats separately and raise the "scalar limit" to 4 for `f32` and `f64`.

eisenwave:
Another consideration is that if you have a `struct` of an `i8` and `i32`, you don't want to bit-pack it so that the second `i32` gets broken up into an `i24` and `i8`. You would want an alignment mechanism so that an `i32` in a `struct` is either passed as `i32` or packed into a surrounding `i64`, but never subdivided.

In other words, bit-pack *every* bit of a small `struct`, including its alignment/padding bits, rather than bit-packing as if it was a `[[gnu::packed]] struct`.

eisenwave:
The wording to get things ideal is a bit tricky; here's a list of test cases I envision:
```cpp
struct { i64 x } -> (i64 x)
struct { i64 x, i64 y } -> (i64 x, i64 y)
struct { i64 x, i64 y, i64 z } -> indirect

struct { i32 x, i32 y } -> (i64 xy)
struct { i32 x, i32 y, i32 z, i32 w } -> (i64 xy, i64 zw)
struct { i8 r, i8 g, i8 b } -> (i32 rgb)
struct { i8 r, i8 g, i8 b, i8 a } -> (i32 rgba)
struct { i8 b, i32 x } -> (i64 bx)
struct { i8 b, i8 c, i32 x } -> (i64 bcx)
struct { i8 b, i16 s, i32 x } -> (i64 bsx)
struct { i16 s, i16 t, i32 x } -> (i64 stx)
struct { i16 s, i16 t, i16 u, i32 x } -> (i64 stu, i32 x)
struct { i16 a, i16 b, i16 c, i16 d } -> (i64 abcd)

struct { f32 x } -> (f32 x)
struct { f32 x, f32 y } -> (f32 x, f32 y)
struct { f32 x, f32 y, f32 z, f32 w } -> (f32 x, f32 y, f32 z, f32 w)
struct { f32 x, f32 y, f32 z, f32 w, f32 v } -> indirect

struct { f64 x } -> (f64 x)
struct { f64 x, f64 y } -> (f64 x, f64 y)
struct { f64 x, f64 y, f64 z, f64 w } -> (f64 x, f64 y, f64 z, f64 w)
struct { f64 x, f64 y, f64 z, f64 w, f64 v } -> indirect

struct { f32 x, i32 y } -> (f32 x, i32 y)
struct { f32 x, i32 y, i32 z } -> (f32 x, i64 yz)
struct { i32 x, f32 y } -> (i32 x, f32 y)
struct { i32 x, i32 y, f32 z } -> (i64 xy, f32 z)
struct { i32 x, i32 y, f32 z, i32 w } -> indirect
```

What strikes me as slightly odd is packing `i32`s into a single `i64`. The packing makes a ton of sense for smaller types like `i8` because they don't really exist in WASM, so packing them into the smallest type that fits them is a somewhat obvious improvement. However, packing separate `i32`s into a single `i64` seems fishy.

This is especially considering that if the WASM host is JavaScript, `i32` can be passed by `number`, but `i64` must be passed by `bigint`, which comes with more overhead.

So in pseudo-code, all that would be:
```python
int_count = 0
int_packed_bits = 0
int_buffer = 0
int_buffer_max = 64
float_count = 0

def flush_buffer():
  if int_packed_bits == 0:
    pass
  elif int_packed_bits <= 32:
    emit_i32(int_packed_bits)
  else:
    emit_i64(int_packed_bits)

for m in get_members_recursively():
  if m.is_integer:
    # Small integers get packed.
    if m.width <= int_buffer_max:
      if int_packed_bits + m.width > int_buffer_max:
        flush_buffer()
      int_buffer |= m.value << int_packed_bits
      int_packed_bits += m.width

    # Large integers (e.g. _BitInt(128)) get split.
    # Huge integers (e.g. _BitInt(256) passed indirectly.
    else:
      for x in m.split_to_i64():
        emit_i64(x)

  if m.is_float:
    flush_buffer()

    # std::float16_t, std::bfloat16_t -> (i32)
    # Never bit-packed together with integers.
    if m.width < 32:
      emit_i32(m.zero_extend_to_i32())

    # std::float32_t -> (f32)
    if m.is_f32:
      emit_f32(x)

    # std::float64_t -> (f64)
    if m.is_f64:
      emit_f64(x)

    # std::float128_t -> (i64, i64)
    if m.is_f128:
      for x in m.split_to_i64():
        emit_i64(x)

flush_buffer()

def emit_i32(x):
  if int_count < 2:
    do_emit("i32", x)
    int_count += 1
  else:
    do_pass_indirectly()

def emit_i64(x):
  if int_count < 2:
    do_emit("i64", x)
    int_count += 1
  else:
    do_pass_indirectly()

def emit_f32(x):
  if float_count < 4:
    do_emit("f32", x)
    float_count += 1
  else:
    do_pass_indirectly()

def emit_f64(x):
  if float_count < 4:
    do_emit("f64", x)
    float_count += 1
  else:
    do_pass_indirectly()
```

